PR #1, opened 2024-07-08
Adding amplitude param

jmfranck:
This is WK's branch from spincore apps -- ported to new combined repo


examples/combined_ODNP.py
```diff
@@ -9,12 +9,12 @@
...
+    hdf_save_dict_to_group,
+)
 import pyspecdata as psd
-from pyspecdata.file_saving.hdf_save_dict_to_group import hdf_save_dict_to_group
-from pyspecdata import strm, DNP_data
-import os
-import sys
-import time
+from pyspecdata import strm
+import os, sys, time
```
jmfranck: these changes didn't make sense, and, in the case of this `DNP_data` probably resulting in non-working code, so I rolled them back

examples/find_gamma_eff.py
```diff
@@ -1,3 +1,4 @@
+# this needs a docstring, which should make it clear why we are doing this
```
jmfranck: this needs a docstring, which should make it clear why you are adding this file

.github/workflows/flake8.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: |
           if [ -n "$CHANGED_FILES" ]; then
             echo "Running Flake8 on changed files: $CHANGED_FILES"
-            flake8 --max-line-length=88 --ignore=E231,E501,E225,E226,E261,E306,E293,W293,E741,W291,E265,E302,E305,E203,W503 $CHANGED_FILES
+            flake8 --max-line-length=88 --ignore=E231,E501,E225,E226,E261,E401,E306,E293,W293,E741,W291,E265,E302,E305,E203,W503 $CHANGED_FILES
```
jmfranck, line 55: multiple imports per line is something we should be cool with

examples/combined_ODNP.py
```diff
@@ -169,6 +169,9 @@
         target_directory = os.path.getcwd()
         filename = "temp_ctrl.h5"
         DNP_data.hdf5_write(filename, directory=target_directory)
+        # this doesn't make any sense -- how
+        # are you trying to write the data
+        # before it has been measured??
```
jmfranck: This doesn't make any sense -- how are you trying to write the data before it has been measured??

examples/combined_ODNP.py
```diff
@@ -24,9 +24,9 @@
 
 final_log = []
 
-logger = init_logging(level="debug")
-target_directory = getDATADIR(exp_type="ODNP_NMR_comp/ODNP")
-fl = figlist_var()
+logger = sys.init_logging(level="debug")
```
jmfranck: this comes from psd, not sys

examples/active.ini
```diff
@@ -1,38 +1,44 @@
...
 deblank_us = 1.0
-acq_time_ms = 1024.0
-nscans = 1
+acq_time_ms = 512.0574266666666
+echo_acq_ms = 5.0
+sw_khz = 3.8999532005615936
+nscans = 2
 thermal_nscans = 3
-carrierfreq_mhz = 14.895116358000001
+carrierfreq_mhz = 14.888852134000002
```
jmfranck: Don't include these types of changes in the PR. Diff against the master branch and only change things you actually mean to change

examples/combined_ODNP.py
jmfranck: Still needs a fix

jmfranck (commented):
Still some things to fix

examples/active.ini
```diff
@@ -30,7 +31,7 @@ num_t1s = 2
 uw_dip_center_ghz = 9.820639
 uw_dip_width_ghz = 0.02
 fir_rep = 500000.0
-min_dBm_step = 1.0
+min_dbm_step = 1.0
```
wfkincaid: IDK if this should be a lower or uppercase 'B'

examples/active.ini
```diff
@@ -1,10 +1,11 @@
 [acq_params]
-amplitude = 1.0
-p90_us = 4.8
+amplitude = 0.575
+p90_us = 1.7
 deadtime_us = 10.0
 tau_us = 3500.0
 deblank_us = 1.0
 acq_time_ms = 1024.0
+sw_khz = 3.8999532005615936
```
wfkincaid, line 8: sw param not included before, do I want to keep this as added?

examples/active.ini
jmfranck: I see -- I don't know if it matters, but switching to lowercase if fine

examples/active.ini
jmfranck: this is OK, actually

jmfranck: b/c you can rationalize that you really want that as a default parameter

examples/combined_ODNP.py
```diff
@@ -168,10 +167,12 @@
         os.remove("temp_ctrl.h5")
         target_directory = os.path.getcwd()
         filename = "temp_ctrl.h5"
-        DNP_data.hdf5_write(filename, directory=target_directory)
-        final_log.append(
-            "change the name accordingly once this is done running!"
-        )
+        control_thermal.hdf5_write(filename, directory=target_directory)
+        # PR this doesn't make any sense -- how
+        # are you trying to write the data
```
jmfranck: once you've dealt with the comments that start with `PR` please remove them

Instruments/nmr_signal_gui.py
```diff
@@ -12,15 +12,16 @@
...
 from .XEPR_eth import xepr as xepr_from_module
-from scipy.interpolate import interp1d
 import time
-import sys, os, random
-import PyQt5 as qt
+import sys
+from PyQt5.QtWidgets import QApplication, QMainWindow, QLabel, QVBoxLayout, QWidget  # QtWidgets is for GUI components
+from PyQt5.QtCore import Qt  # QtCore is for core non-GUI functionalities
+from PyQt5.QtGui import QIcon  # QtGui is for handling icons and images
+from PyQt5.QtWidgets import QFileDialog, QMessageBox, QLineEdit, QComboBox, QPushButton, QCheckBox, QSlider, QHBoxLayout, QAction
```
jmfranck, line 20: to get these, I removed the import * line, then copy/pasted the output of flake into chatgpt to ask for the import statements.

I confirm that I can run these lines in ipython, so I think that means that these are all found in the right place

Instruments/nmr_signal_gui.py
```diff
@@ -168,6 +169,7 @@ def adc_offset(self):
         print("adc was ", self.myconfig["adc_offset"], end=" and ")
         counter = 0
         first = True
+        result1 = result2 = result3 = None
```
jmfranck, line 172: this is just to make flake happy

jmfranck (commented):
I make a couple comments for you to see, so you know what's up, and then you need to remove the `#PR ` statements that were left in, if they have been dealt with

examples/find_gamma_eff.py
```diff
@@ -0,0 +1,14 @@
+# PR this needs a docstring, which should make it clear why we are doing this
```
jmfranck, line 11: This comment was not dealt with, actually -- please supply the docstring.  I actually am not sure what this is for.

jmfranck (commented):
aside from the # PR that were left in, and the things I note for review, this is OK

examples/active.ini
guinneak198, line 8: There is no default for this parameter so it needs to be defined in the active ini

examples/combined_ODNP.py
```diff
@@ -218,13 +214,9 @@
 vd_data.set_prop("postproc_type", IR_postproc)
 nodename = vd_data.name()
 # {{{ again, implement a file fallback
-with h5py.File(
-    os.path.normpath(os.path.join(target_directory, f"{filename}"))
-) as fp:
+with h5py.File(os.path.normpath(os.path.join(target_directory, f"{filename}"))) as fp:
```
jmfranck, line 217: very soon, please make a new PR to implement your new function.  Note that you will also want to pass it the log info as a kwarg.

examples/active.ini
wfkincaid: the NMR signal GUI also needs 'adc_offset' defined, should that also be included?

examples/find_gamma_eff.py
wfkincaid: When not being able to run the GUI, this is used for getting on resonance

examples/combined_ODNP.py
wfkincaid: AG showed me an example so that I can do this

examples/active.ini
jmfranck: I don't think that's true?

The point is that you have to press the button that determines the ADC offset before trying to acquire signal.

please reply here to confirm this, so that we can can resolve this

examples/combined_ODNP.py
jmfranck: there was something added to the file, which I deleted.  that would have been doubled up (if it worked) and we also want to do this in a separate PR.

jmfranck: @wfkincaid please don't resolve conversations -- I use resolving them as a way to know which issues are taken care of, and if I can trust others haven't resolved them, it moves much faster

examples/active.ini
jmfranck: so, what I'm saying is that `adc_offset` actually *should not* be in this file, b/c if you try to acquire NMR without first clicking the ADC offset button, the NMR GUI *should* fail.

guinneak198: yes JF is correct if the GUI or anything is run without the adc_offset being set (which is set by running adc_offset.py) an error is thrown saying you skipped a step and need to determine adc_offset

examples/find_gamma_eff.py
guinneak198: added updated docstring

examples/active.ini
jmfranck: I'm going to go ahead and merge this PR.. please make sure that @wfkincaid is on the same page with this issue

PR #4, opened 2024-07-15
fix error with p90 array

guinneak198:
not sure what heppened but when I run FLInst NMRsignal on master I get this error ```
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\scripts\FLInst-script.py", line 11, in <module>
    load_entry_point('francklab-instruments', 'console_scripts', 'FLInst')()
  File "c:\apps-su\flinst\Instruments\cmd.py", line 27, in cmd
    cmds[sys.argv[1]]()
  File "c:\apps-su\flinst\Instruments\nmr_signal_gui.py", line 442, in main
    tunwin = NMRWindow(x, myconfig)
  File "c:\apps-su\flinst\Instruments\nmr_signal_gui.py", line 43, in __init__
    self.acq_NMR()
  File "c:\apps-su\flinst\Instruments\nmr_signal_gui.py", line 187, in acq_NMR
    self.generate_data()
  File "c:\apps-su\flinst\Instruments\nmr_signal_gui.py", line 149, in generate_data
    ret_data=None,
  File "c:\apps-su\flinst\SpinCore_pp\ppg\echo.py", line 123, in run_spin_echo
    ("delay", repetition_us),
  File "c:\apps-su\flinst\SpinCore_pp\SpinCore_pp.py", line 208, in load
    ppg_element(*a_tuple)
TypeError: in method 'ppg_element', argument 2 of type 'double'
```
After investigation I find that the p90 and p180 we feed to the SpinCore.load is an array and not a float

### DIFF
--- a/SpinCore_pp/ppg/echo.py
+++ b/SpinCore_pp/ppg/echo.py
@@ -112,10 +112,10 @@ def run_spin_echo(
             [
                 ("phase_reset", 1),
                 ("delay_TTL", deblank_us),
-                ("pulse_TTL", prog_p90_us.item(), "ph1", ph1_cyc),
+                ("pulse_TTL", prog_p90_us, "ph1", ph1_cyc),
                 ("delay", tau_us),
                 ("delay_TTL", deblank_us),
-                ("pulse_TTL", prog_p180_us.item(), "ph2", ph2_cyc),
+                ("pulse_TTL", prog_p180_us, "ph2", ph2_cyc),
                 ("delay", deadtime_us),
                 ("acquire", acq_time_ms),
                 ("delay", repetition_us),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: SpinCore_pp/ppg/echo.py [hunk 1]
```diff
@@ -112,10 +112,10 @@ def run_spin_echo(
             [
                 ("phase_reset", 1),
                 ("delay_TTL", deblank_us),
-                ("pulse_TTL", prog_p90_us, "ph1", ph1_cyc),
+                ("pulse_TTL", prog_p90_us.item(), "ph1", ph1_cyc),
                 ("delay", tau_us),
                 ("delay_TTL", deblank_us),
-                ("pulse_TTL", prog_p180_us, "ph2", ph2_cyc),
+                ("pulse_TTL", prog_p180_us.item(), "ph2", ph2_cyc),
```
jmfranck: Don't do this here. Fix in the function that determines the programmed pulse length. If the length of the returned value is one, then call item. 

This has to do with some recent updates that I made to extend the range of applicability of that function

jmfranck (changes requested):

PR #5, opened 2024-07-15
Using CH3 of GDS

wfkincaid:


jmfranck:
> I'm not sure what this was attempting to do. It is very large, and I think it was trying to merge Master into a different branch, which is odd because you can do that yourself. Also, I thought we had already merged the ability to do channel 3 in at some point

PR #9, opened 2024-07-15
MWTune gui

wfkincaid:


Instruments/microwave_tuning_gui.py
```diff
@@ -88,7 +88,7 @@ def orig_zoom_limits(self):
         # a plot with the dip centered
         #
         # dip shown on slack (https://jmfrancklab.slack.com/archives/CLMMYDD98/p1705090019740609)
-        for ini_val, w in [("9816000", self.textbox1), ("9823000", self.textbox2)]:
+        for ini_val, w in [("9703000", self.textbox1), ("9710000", self.textbox2)]:
```
jmfranck: This is a hard-coded change, but we had discussed making this adjustable

jmfranck (changes requested):

Instruments/microwave_tuning_gui.py
```diff
@@ -200,7 +210,6 @@ def regen_plots(self):
             # {{{ for the last trace, interpolate, and find the min
             if hasattr(self, "interpdata"):
                 xx, yy = self.interpdata
-                dip_frq_GHz = self.dip_frq_GHz
```
jmfranck, line 203: this was the line causing it to fail flake.  The variable on the left is genuinely never used.

jmfranck (approved):

PR #10, opened 2024-07-17
WK debug MW tuning gui to find dip using ini file, works for new probe

wfkincaid:


Instruments/microwave_tuning_gui.py
```diff
@@ -13,28 +13,20 @@
...
 import PyQt5.QtWidgets as qt5w
 import SpinCore_pp  # just for config file, but whatever...
-
-from matplotlib.backends.backend_qt5agg import (
-    FigureCanvasQTAgg as FigureCanvas,
-)
-from matplotlib.backends.backend_qt5agg import (
-    NavigationToolbar2QT as NavigationToolbar,
-)
+import matplotlib.backends.backend_qt5agg as mqt5ag
```
jmfranck, line 19: I matched your import style above

Instruments/microwave_tuning_gui.py
```diff
@@ -92,17 +84,17 @@ def orig_zoom_limits(self):
...
-            ),
-            (
-                (myconfig["uw_dip_center_ghz"] + myconfig["uw_dip_width_ghz"]),
-                self.textbox2,
-            ),
-        ]:
+        # dip shown on slack (https://jmfrancklab.slack.com/archivec/CLMMYDD98/p1705090019740609)
+        for ini_val, w in zip(
+            [
+                str(int((myconfig["uw_dip_center_ghz"] + j) * 1e6))
```
jmfranck, line 90: rather than wrapping both with these conversion functions, can do it once inside a list comprehension

PR #11, opened 2024-07-22
Fid

guinneak198:
ppg for single pulse NUTATION!


examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+the FID with a 4-step phase cycle. 
+"""
+
+from pyspecdata import getDATADIR
+from numpy import r_
+import sys
+import os
+import SpinCore_pp
+from SpinCore_pp import prog_plen, get_integer_sampling_intervals, save_data
+from SpinCore_pp.ppg import generic
```
jmfranck: please do a `import SpinCore_pp as ???` instead

examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+    nPoints,
+    config_dict["SW_kHz"],
+    config_dict["acq_time_ms"],
+) = get_integer_sampling_intervals(
+    SW_kHz=config_dict["SW_kHz"],
+    time_per_segment_ms=config_dict["acq_time_ms"],
+)
+# }}}
+# {{{add file saving parameters to config dict
+config_dict["type"] = "FID"
```
jmfranck: FID_nutation

examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["echo_counter"] += 1
```
jmfranck: FID_nutation_counter

examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+    print(
+        "Based on that, and the gamma_eff_MHz_G you have in your .ini file, I'm setting the field to %f"
+        % field_G
+    )
+    with xepr() as x:
+        assert field_G < 3700, "are you crazy??? field is too high!"
+        assert field_G > 3300, "are you crazy?? field is too low!"
+        field_G = x.set_field(field_G)
+        print("field set to ", field_G)
+# }}}
```
jmfranck: can we please wrap this into a function?

examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+prog_p90_us = prog_plen(config_dict["p90_us"])
+# {{{check total points
+total_pts = nPoints * nPhaseSteps
+assert total_pts < 2**14, (
+    "You are trying to acquire %d points (too many points) -- either change SW or acq time so nPoints x nPhaseSteps is less than 16384"
+    % total_pts
+)
+# }}}
+# {{{ acquire FID
```
jmfranck: where is the loop to run a nutation curve?

examples/run_onepulse.py
```diff
@@ -0,0 +1,101 @@
...
+    time_per_segment_ms=config_dict["acq_time_ms"],
+    SW_kHz=config_dict["SW_kHz"],
+    amplitude=config_dict["amplitude"],
+    ret_data=None,
+)
+# }}}
+# {{{ chunk and save data
+data.chunk("t", ["ph1", "t2"], [len(ph1_cyc), -1])
+data.setaxis("ph1", ph1_cyc / 4)
+data.set_prop("postproc_type", "spincore_FID_v1")
```
jmfranck: no, we want a nutation curve

jmfranck (changes requested):

examples/run_onepulse.py
guinneak198: added to simple_fns

examples/run_onepulse.py
guinneak198: fixed

examples/run_onepulse.py
guinneak198: fixed

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+# }}}
+input(
+    "I'm assuming that you've tuned your probe to %f since that's what's in your .ini file. Hit enter if this is true"
+    % config_dict["carrierFreq_MHz"]
+)
+# {{{ let computer set field
+if adjust_field:
+    field_G = config_dict["carrierFreq_MHz"] / config_dict["gamma_eff_MHz_G"]
+    spc.set_field(Field_G=config_dict["carrierFreq_MHz"] / config_dict["gamma_eff_MHz_G"])
+# }}}
```
jmfranck: move all this into your set field routine.  Pass the set field routine sys.argv as an argument.

SpinCore_pp/simple_fns.py
```diff
@@ -26,3 +29,23 @@ def get_integer_sampling_intervals(SW_kHz, time_per_segment_ms):
...
+    with xepr() as x:
+        assert Field_G < 3700, (
+            "Are you mad?? The field you want, %g, is too high!" % Field_G
+        )
+        assert Field_G > 3300, (
+            "Are you mad?? The field you want, %g, is too low!" % Field_G
+        )
+        Field_G = x.set_field(Field_G)
+        print("Field set to ", Field_G)
+    return
```
jmfranck: this is not a "simple function" -- put it in its own module?

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,102 @@
...
+        nPoints=nPoints,
+        time_per_segment_ms=config_dict["acq_time_ms"],
+        SW_kHz=config_dict["SW_kHz"],
+        amplitude=config_dict["amplitude"],
+        ret_data=data,
+    )
+# }}}
+data.rename("indirect", "p_90")
+data.setaxis("p_90", prog_p90_us * 1e-6).set_units("p_90", "s")
+data.set_prop("desired_p90s", p90_range_us)
```
jmfranck: these variable names confuse me, and we may want to note that this could be a problem elsewhere:

here, you want to set the axis to `p90_range_us`, which are the effective pulse lengths that you were shooting for.  The `prog_p90_us` are the values that you needed to ask spincore for to get those.

jmfranck (changes requested):
I think that these should mostly cover it.

examples/run_onepulse.py
guinneak198: made separate module in field_control instead

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,24 @@
...
+        Field_G = config_dict["carrierFreq_MHz"] / config_dict["gamma_eff_MHz_G"]
+        with xepr() as x:
+            assert Field_G < 3700, (
+                "Are you mad?? The field you want, %g, is too high!" % Field_G
+            )
+            assert Field_G > 3300, (
+                "Are you mad?? The field you want, %g, is too low!" % Field_G
+            )
+            Field_G = x.set_field(Field_G)
+            print("Field set to ", Field_G)
```
jmfranck: Logger info

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,91 @@
...
+import sys
+import os
+import SpinCore_pp
+import SpinCore_pp as spc
+from SpinCore_pp.ppg import generic
+from datetime import datetime
+
+my_exp_type = "ODNP_NMR_comp/nutation"
+assert os.path.exists(getDATADIR(exp_type=my_exp_type))
+desired_p90_range_us = linspace(0.5, 15, 20, endpoint=False)
```
jmfranck: Don't do this endpoint false thing.

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,91 @@
...
+    config_dict["acq_time_ms"],
+) = spc.get_integer_sampling_intervals(
+    SW_kHz=config_dict["SW_kHz"],
+    time_per_segment_ms=config_dict["acq_time_ms"],
+)
+# }}}
+# {{{add file saving parameters to config dict
+config_dict["type"] = "FID_nutation"
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["FID_nutation_counter"] += 1
```
jmfranck, line 35: Not for this pull request, but if you're ambitious next time this comes up: I've realized that the way we handle counters is ridiculous.  We should just modify the config parser so that it automatically keeps track of a counter for each "type".  Then we keep all the counters in their own section, and have a special function to increment the counter for each types. It should not be required to modify every time we want a counter with a new name.

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,24 @@
+from Instruments.XEPR_eth import xepr
+
+
+def set_field(argv, config_dict):
+    """Ensures the field is appropriate before setting the desired field."""
+    if len(argv) == 2 and argv[1] == "stayput":
```
jmfranck: I think do this outside the function.  It's weird to pass `says.argv`, since it's so specific to the script.  So move this to the script, and call the function inside the if block.

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,91 @@
...
+# }}}
+# {{{ command-line option to leave the field untouched (if you set it once, why set it again)
+spc.set_field(sys.argv, config_dict)
+# }}}
+# {{{set phase cycling
+ph1_cyc = r_[0, 1, 2, 3]
+nPhaseSteps = 4
+# }}}
+prog_p90_us = spc.prog_plen(desired_p90_range_us)
```
jmfranck: This function now takes beta and amplitude, right?  If not, we need to merge that PR first.

Because of this, I would multiply the desired p90 range here by sqrt expected power (to get a desired beta)

jmfranck (changes requested):

examples/run_onepulse_nutation.py
jmfranck: Literally right here, on this line do the mult to keep it simple.  Also pass amp as second arg, obviously

examples/run_onepulse_nutation.py
guinneak198: I agree on the idea here but also agree to save for another PR

examples/run_onepulse_nutation.py
guinneak198: updated to use beta

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,23 @@
+from Instruments.XEPR_eth import xepr
+import pyspecdata as psd
+
+logger = psd.init_logging(level="info")
```
jmfranck: This is not a script. Do not set the logging level

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,23 @@
...
+
+def set_field(config_dict):
+    """Ensures the field is appropriate before setting the desired field."""
+    input(
+        "I'm assuming that you've tuned your probe to %f since that's what's in your .ini file. Hit enter if this is true"
+        % config_dict["carrierFreq_MHz"]
+    )
```
jmfranck: This is not part of setting the field, and should go in the pulse sequence... The point of this is to make sure that the probe is tuned to the frequency you're going to be sending high power pulses at

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,23 @@
+from Instruments.XEPR_eth import xepr
+import pyspecdata as psd
+
+logger = psd.init_logging(level="info")
+
+def set_field(config_dict):
+    """Ensures the field is appropriate before setting the desired field."""
```
jmfranck: The key point is that the field is set based on the carrier frequency and the estimated gamma

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,94 @@
...
+    time_per_segment_ms=config_dict["acq_time_ms"],
+)
+# }}}
+# {{{add file saving parameters to config dict
+config_dict["type"] = "FID_nutation"
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["FID_nutation_counter"] += 1
+# }}}
+# {{{ command-line option to leave the field untouched (if you set it once, why set it again)
+if len(sys.argv) == 2 and sys.argv[1] == "stayput":
```
jmfranck, line 42: In keeping with my other comment above, you should be checking for the tuning frequency in the line above here, or or further up or closer the actual pulsing whatever makes sense for you

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,94 @@
...
+        nPoints=nPoints,
+        time_per_segment_ms=config_dict["acq_time_ms"],
+        SW_kHz=config_dict["SW_kHz"],
+        amplitude=config_dict["amplitude"],
+        ret_data=data,
+    )
+# }}}
+data.rename("indirect", "beta")
+data.setaxis("beta", beta_range_s_sqrtW).set_units("beta", "s√W")
+data.set_prop("prog_p90s", prog_p90_us)
```
jmfranck: The property should be named the same thing as the variable.... In any event, you need to include the _us

jmfranck: Unless you're going to convert it to seconds

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,94 @@
...
+# {{{ chunk and save data
+data.chunk("t", ["ph1", "t2"], [len(ph1_cyc), -1])
+data.setaxis("ph1", ph1_cyc / 4)
+if config_dict["nScans"] > 1:
+    data.setaxis("nScans", r_[0 : config_dict["nScans"]])
+data.reorder(["nScans", "ph1", "beta", "t2"])
+data.set_units("t2", "s")
+data.set_prop("postproc_type", "spincore_FID_nutation_v1")
```
jmfranck: You need to increment the version number when making these other changes

jmfranck (changes requested):

examples/run_onepulse_nutation.py
guinneak198: good catch

examples/run_onepulse_nutation.py
guinneak198: This is the first version of FID nutation

guinneak198: falls in line with this [PR](https://github.com/jmfrancklab/proc_scripts/pull/110) in proc_scripts

jmfranck: there are other changes being made here.  Increment the version number.

SpinCore_pp/field_control.py
```diff
@@ -0,0 +1,19 @@
...
+    Field_G = config_dict["carrierFreq_MHz"] / config_dict["gamma_eff_MHz_G"]
+    with xepr() as x:
+        assert Field_G < 3700, (
+            "Are you mad?? The field you want, %g, is too high!" % Field_G
+        )
+        assert Field_G > 3300, (
+            "Are you mad?? The field you want, %g, is too low!" % Field_G
+        )
+        Field_G = x.set_field(Field_G)
+        psd.logger.info(psd.strm("Field set to ", Field_G))
```
jmfranck: you should not get logger from psd -- do import logging and `logging.info`

examples/run_onepulse_nutation.py
```diff
@@ -0,0 +1,98 @@
...
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["FID_nutation_counter"] += 1
+# }}}
+# {{{ command-line option to leave the field untouched (if you set it once, why set it again)
+if len(sys.argv) == 2 and sys.argv[1] == "stayput":
+    pass
+else:
+    input(
+        "I'm assuming that you've tuned your probe to %f since that's what's in your .ini file. Hit enter if this is true"
+        % config_dict["carrierFreq_MHz"]
```
jmfranck: again, think about what this is for --> even if you are staying put, you still want to be sure that you are tuned!

examples/run_onepulse_nutation.py
jmfranck: you have not changed the name of the property!

jmfranck (changes requested):
once these changes are dealt with, we should be good to go

examples/run_onepulse_nutation.py
```diff
@@ -17,7 +17,7 @@
 
 my_exp_type = "ODNP_NMR_comp/nutation"
 assert os.path.exists(getDATADIR(exp_type=my_exp_type))
-beta_range_s_sqrtW = linspace(0.1e-6, 150e-6, 20)
+beta_s_sqrtW_array = linspace(0.1e-6, 150e-6, 20)
```
jmfranck, line 20: better variable name

examples/run_onepulse_nutation.py
```diff
@@ -81,13 +81,13 @@
...
-data.setaxis("beta", beta_range_s_sqrtW).set_units("beta", "s√W")
-data.set_prop("prog_p90_us", prog_p90_us)
+data.setaxis("beta", beta_s_sqrtW_array).set_units("beta", "s√W")
+data.set_prop("prog_p90_us", p90_us_array)
 # {{{ chunk and save data
 data.chunk("t", ["ph1", "t2"], [len(ph1_cyc), -1])
 data.setaxis("ph1", ph1_cyc / 4)
 if config_dict["nScans"] > 1:
-    data.setaxis("nScans", r_[0 : config_dict["nScans"]])
+    data.setaxis("nScans", "#")
```
jmfranck: this does the same thing + is shorter

jmfranck (approved):

PR #12, opened 2024-07-29
calibrated pulse

guinneak198:
ppg to acquire pulse captures on the gdss

### DIFF
--- a/SpinCore_pp/beta_conv.py
+++ b/SpinCore_pp/beta_conv.py
@@ -1,0 +1,78 @@
+from pylab import array
+import numpy as np
+from pyspecdata import r_, nddata
+
+
+def prog_beta(desired_actual):
+    """
+    Takes the desired beta (us *sqrt(W)) and tells the
+    user what beta should be programmed in order to get the actual desired
+    beta
+    Parameters
+    ==========
+    desired_actual: float
+                    the actual beta you wish the spincore to output,
+                    in us sqrt(W)
+    Returns
+    =======
+    retval: float
+            The beta you tell spincore in order to get the desired actual.
+    """
+    # {{{ list of programmed beta, and actual beta - used in
+    # generating the calibrated fit
+    datapoints = [
+        (1, 2.25869),
+        (2, 5.78065),
+        (3, 10.4132),
+        (3.5, 13.2053),
+        (4, 16.9808),
+        (4.5, 20.4275),
+        (5, 24.8538),
+        (5.5, 30.0159),
+        (6, 35.6903),
+        (6.5, 42.2129),
+        (7, 48.9577),
+        (7.5, 55.977),
+        (8, 63.0445),
+        (8.5, 69.881),
+        (9, 76.9727),
+        (9.5, 84.51),
+        (10, 91.700),
+        (10.5, 98.63),
+        (11, 105.327),
+        (11.5, 112.9),
+        (12, 120.264),
+        (12.5, 126.2),
+    ]
+    # neat JF trick for organizing these data points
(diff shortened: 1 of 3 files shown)


SpinCore_pp/config_parser_fn.py
```diff
@@ -231,6 +230,12 @@ class configuration(object):
             0,
             "number of field sweeps performed for a particular sample that day",
         ),
+        "misc_counter": (
+            int,
+            "file_names",
+            0,
+            "number of miscellaneous expts performed for a particular sample that day",
+        ),
```
guinneak198, line 238: rather than making a counter specifically for calibrating pulses that will maybe be used once a year I make this a misc counter so that even with test equipment scripts we can use the active.ini

examples/calib_pulses.py
```diff
@@ -0,0 +1,86 @@
...
+    gds.write(":CHAN2:DISP ON")
+    gds.write(":CHAN3:DISP OFF")
+    gds.write(":CHAN4:DISP OFF")
+    gds.CH2.voltscal = 100e-3  # set voltscale to 100 mV
+    gds.timescal(20e-6, pos=0)  # set timescale to 20 us
+    gds.write(":TIM:MOD WIND")
+    gds.write(":CHAN2:IMP 5.0E+1")  # set impedance to 50 ohm
+    gds.write(":TRIG:SOUR CH2")
+    gds.write(":TRIG:MOD NORMAL")  # set trigger mode to normal
+    gds.write(":TRIG:HLEV 7.5E-2")  # used in gds_for_tune which seems reasonable
```
guinneak198, line 44: initial settings on gds - needs to be tested to get optimal settings (these are rough guesses)

SpinCore_pp/beta_conv.py
```diff
@@ -0,0 +1,78 @@
+from pylab import array
+import numpy as np
+from pyspecdata import r_, nddata
+
```
guinneak198, line 4: most of this is all taken/borrowed from pulse_length_conv.py but just exchanging p90 for beta

SpinCore_pp/beta_conv.py
```diff
@@ -0,0 +1,78 @@
...
+        (8, 63.0445),
+        (8.5, 69.881),
+        (9, 76.9727),
+        (9.5, 84.51),
+        (10, 91.700),
+        (10.5, 98.63),
+        (11, 105.327),
+        (11.5, 112.9),
+        (12, 120.264),
+        (12.5, 126.2),
```
guinneak198, line 45: these will be updated based on the output of the example script to capture the pulses

examples/calib_pulses.py
```diff
@@ -0,0 +1,124 @@
...
+    nPoints,
+    config_dict["SW_kHz"],
+    config_dict["acq_time_ms"],
+) = spc.get_integer_sampling_intervals(
+    config_dict["SW_kHz"], config_dict["acq_time_ms"]
+)
+# }}}
+sqrt_P = config_dict["amplitude"] * np.sqrt(75)  # we have a 75 W amplifier
+prog_beta = np.linspace(10, 70, 50, endpoint=False)
+prog_p90s = prog_beta / sqrt_P
```
guinneak198, line 22: makes list of betas and the p90s based on the assumption that the sqrt(P) = amplitude * sqrt(75)

jmfranck:
It seems like you are creating a new function in a new file where you are just copying code from the existing pulse calibration. The existing pulse calibration function should not change much, and should therefore just be an edit of the existing file. If you think it's good to change the name of the function, that's fine

PR #13, opened 2024-07-29
Calib pulse length

guinneak198:
there is no existing example to acquire the pulse lengths as captured on the gds


jmfranck:
There is already an example called run nutation amplitude, previously discussed, which appears to be doing the same thing that you're trying to do here. Start by modifying that. Also, do not do data processing in the same script where you acquire the data

PR #14, opened 2024-07-29
Calib pulse length

guinneak198:
acquire capture of pulses based on setting of beta value to calibrate pulses


SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,97 @@
...
+        (17.2001, 3.5688, 10.3898),
+        (18.9997, 4.0651, 12.4032),
+        (20.8002, 4.5997, 14.5975),
+        (22.5998, 5.2121, 17.1804),
+        (24.40027, 5.8534, 20.1967),
+        (26.19987, 6.5583, 23.2631),
+        (28.0003, 7.2702, 26.7712),
+        (29.7999, 8.0251, 30.3686),
+        (31.6004, 8.9074, 34.5012),
+        (33.4, 9.7384, 38.7034),
```
guinneak198: new values acquired today where the first value is beta = `linspace(10,100,50)` and I then convert to a 90 time to tell the spincore by dividing by sqrt(75) (75 W amplifier). The second value is ∫√(P(t))dt of the 90 pulse capture and the third value is ∫√(P(t))dt for the 180 pulse capture

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,97 @@
...
     plen_prog = r_[0, prog90, 2 * prog90]
     # assume the longest pulse is about the correct length
     # and again gather into an array
-    plen_actual = r_[0, act90, act180] * 2 * prog90[-1] / act180[-1]
+    plen_actual = r_[0, beta90, beta180] * 2 * prog90[-1] / beta180[-1]
 
     # }}}
     def zonefit(desired_actual):
-        if desired_actual < 25:
+        if desired_actual < 100:
```
guinneak198: scaled this since now we are dealing with beta and not p90 (the upper limit of acquisition was a beta = 100)

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,97 @@
...
             mask = np.ones_like(plen_prog, dtype=bool)
         else:
-            mask = plen_prog > 10
+            mask = plen_prog > 40
```
guinneak198: not sure how 10 was decided but since above I changed 25 to 100 I just multiplied 10 by 4 as well

examples/90time_conv.py
```diff
@@ -12,10 +12,11 @@
 from SpinCore_pp.pulse_length_conv import prog_plen
 
 # the following choice fails dramatically with the old code
-desired_plen = r_[0:50:50j]
+desired_plen = r_[10:150:50j]
```
guinneak198: no longer dealing with pulse times but rather beta values

examples/calib_pulses.py
```diff
@@ -0,0 +1,71 @@
...
+)
+# }}}
+# {{{ add file saving parameters to config dict
+config_dict["type"] = "pulse_calib"
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["misc_counter"] += 1
+# }}}
+# {{{ ppg
+sqrt_P = config_dict["amplitude"] * np.sqrt(75)  # we have a 75 W amplifier
+prog_beta = np.linspace(10, 100, 50, endpoint=False)
```
guinneak198: I want to acquire linearly spaced beta!

guinneak198: units of us*sqrt(W)

examples/calib_pulses.py
```diff
@@ -0,0 +1,71 @@
...
+p90s_range = prog_beta / sqrt_P
```
guinneak198: now calculate a 90 time to send spincore based off the desired betas and the calculation of sqrt(P)

examples/calib_pulses.py
```diff
@@ -0,0 +1,71 @@
...
+                ("delay",config_dict["tau_us"]),
+                ("delay_TTL", 1.0),
+                ("pulse_TTL", p180, 0),
+                ("delay", config_dict["deadtime_us"]),
+                ("acquire", config_dict["acq_time_ms"]),
+            ]
+        )
+        spc.stop_ppg()
+        spc.runBoard()
+        datalist.append(gds.waveform(ch=2))
```
guinneak198: because I am dealing with a calibration I believe it's more straightforward and less error prone if I don't use premade ppg functions and rather explicitly state what I am doing

examples/calib_pulses.py
```diff
@@ -0,0 +1,71 @@
...
+        spc.stopBoard()
+data = psd.concat(datalist, "p_90").reorder("t")        
+data.set_units("t", "s")
+data.set_prop("set_p90s", p90s_range)
+data.set_prop("set_p180s", 2*p90s_range)
+data.set_prop("set_beta", prog_beta)
+data.set_prop("acq_params", config_dict.asdict())
```
guinneak198: when calibrating/processing this will be a nice record of the values used

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,97 @@
...
     desired_actual: float
-                    the actual pulse length you wish the spincore to output,
-                    in us
+                    the actual beta you wish the spincore to output,
+                    in us*sqrt(W)
     Returns
     =======
     retval: float
-            The pulse length you tell spincore in order to get the desired actual.
+            The beta you tell spincore in order to get the desired actual.
```
jmfranck: you don't tell spincore to use a particular value of beta -- you tell it to use a pulse length

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,97 @@
 
 def prog_plen(desired_actual):
```
jmfranck: you probably want an amplitude here as well

SpinCore_pp/pulse_length_conv.py
jmfranck: this is the region where it's curved

SpinCore_pp/__init__.py
```diff
@@ -12,7 +12,7 @@
     tune,
     adc_offset,
 )
-from .pulse_length_conv import prog_plen
+from .pulse_length_conv import prog_plen,prog_plen_lo
```
jmfranck: what is this? why a new function?

SpinCore_pp/__init__.py
```diff
@@ -33,6 +33,7 @@
     "ppg_element",
     "process_args",
     "prog_plen",
+    "prog_plen_lo",
```
jmfranck: see above

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,195 @@
...
+        (89.1998, 64.9480, 168.1456),
+        (91.0002, 67.2708, 172.1686),
+        (92.7998, 69.4676, 176.1556),
+        (94.6003, 71.6768, 180.6902),
+        (96.3999, 73.8858, 184.8543),
+        (98.2004, 75.7488, 188.7464),
     ]
     # neat JF trick for organizing these data points
-    prog90, act90, act180 = map(array, zip(*datapoints))
+    prog_beta90, beta90, beta180 = map(array, zip(*datapoints))
```
jmfranck: This doesn't make sense.  There should be two arrays here -- the programmed pulse length, and the measured value of beta.  You can just call these `t_p` and `beta`.

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,195 @@
...
     # }}}
     # {{{ prepare data into arrays for interpolation
     # gather programmed pulse lengths in array
-    plen_prog = r_[0, prog90, 2 * prog90]
+    plen_prog = r_[0, prog_beta90, 2 * prog_beta90]
     # assume the longest pulse is about the correct length
     # and again gather into an array
-    plen_actual = r_[0, act90, act180] * 2 * prog90[-1] / act180[-1]
+    plen_actual = r_[0, beta90, beta180] * 2 * prog_beta90[-1] / beta180[-1]
```
jmfranck: this is all affected by above.

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,195 @@
...
+            c = calibration_data.polyfit("plen", order=10)
+        else:
+            c = calibration_data.polyfit("plen", order=1)
+        return np.polyval(c[::-1], desired_actual)
+    ret_val = np.vectorize(zonefit)(desired_actual)
+    if ret_val.size > 1:
+        return ret_val
+    else:
+        return ret_val.item()
+def prog_plen_lo(desired_actual):
```
jmfranck: you should not have a separate function.  You should just have an argument to the first function, and choose the different calibration table based on the amplitude setting.

SpinCore_pp/pulse_length_conv.py
```diff
@@ -5,69 +5,195 @@
...
+        (81.9996, 56.7011, 152.1547),
+        (83.8001, 58.7643, 155.8218),
+        (85.5997, 60.8571, 159.8310),
+        (87.4002, 62.9501, 164.1420),
+        (89.1998, 64.9480, 168.1456),
+        (91.0002, 67.2708, 172.1686),
+        (92.7998, 69.4676, 176.1556),
+        (94.6003, 71.6768, 180.6902),
+        (96.3999, 73.8858, 184.8543),
+        (98.2004, 75.7488, 188.7464),
```
jmfranck: I would note that (I think this was my initial approach, so not blaming here :-) ) this may not be the cleanest approach.

Note that the only info that we really need are the 10 polyfit coefficients, the linear threshold, and the linear coefficients that are used below.

You could think of determining and validating these in your processing script, and then just using them here.  That might clarify what's going wrong with the calibraqtion.

jmfranck (changes requested):

SpinCore_pp/pulse_length_conv.py
jmfranck: specifically note that in the code where you process the data that you acquired, you should have β as a function of t_p.  It's important to note that here, we have a polynomial that gives t_p as a function of β.  You can flip as follows:
```
# assume d gives the values of beta on a coordinate axis of t_p
d_flipped = d.shape.rename('t_p','beta').alloc()
d_flipped['beta'] = d.data
d_flipped.data = d['t_p'].copy()
# now you can call polyfit w/ d_flipped, etc.
```

examples/calib_pulses.py
```diff
@@ -0,0 +1,81 @@
...
+sqrt_P = config_dict["amplitude"] * np.sqrt(75)  # we have a 75 W amplifier
+prog_beta = np.linspace(10, 100, 50, endpoint=False)
+p90s_range = prog_beta / sqrt_P
+tx_phases = np.r_[0.0, 90.0, 180.0, 270.0]
+Rx_scans = 1
+datalist = []
+with GDS_scope() as gds:
+    for index, val in enumerate(p90s_range):
+        p90 = val
+        p180 = 2 * val
```
jmfranck: there should be no "p180" just call `val` `t_p` and remove the next two lines.  Do not call this p90 it is not a 90 degree pulse most or all of the time!

examples/calib_pulses.py
```diff
@@ -0,0 +1,81 @@
...
+        spc.load(
+            [
+                ("phase_reset", 1),
+                ("delay_TTL", 1.0),
+                ("pulse_TTL", p90, 0),
+                ("delay", config_dict["tau_us"]),
+                ("delay_TTL", 1.0),
+                ("pulse_TTL", p180, 0),
+                ("delay", config_dict["deadtime_us"]),
+                ("acquire", config_dict["acq_time_ms"]),
```
jmfranck: only one pulse, and no need to acquire here!

examples/calib_pulses.py
```diff
@@ -0,0 +1,81 @@
...
+# }}}
+# {{{ ppg
+sqrt_P = config_dict["amplitude"] * np.sqrt(75)  # we have a 75 W amplifier
+prog_beta = np.linspace(10, 100, 50, endpoint=False)
+p90s_range = prog_beta / sqrt_P
+tx_phases = np.r_[0.0, 90.0, 180.0, 270.0]
+Rx_scans = 1
+datalist = []
+with GDS_scope() as gds:
+    for index, val in enumerate(p90s_range):
```
jmfranck: in between these two lines, you need to do your scope settings.  This is where AB's scope-acquired NMR programs should probe helpful.

examples/calib_pulses.py
jmfranck: in general, please don't put comments in the PR that belong as code comments.

But also, this is not true.  You just want to acquire what the β value is for a set of linearly spaced pulse lengths.

You do want to use the amplitude to help you determine the longest pulse length that you're going to use -- using similar math to what you're using here, which would be smart.

You always want to start from something like 0.5 μs so that you can really see the nonlinearity that you're trying to calibrate for.

I do not understand why you're passing `endpoint=False` to linspace here.

jmfranck (commented):
here I review the program that actually does the acquisition

SpinCore_pp/__init__.py
guinneak198, line 15: new calibration table for low amplitude

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
 from pyspecdata import r_, nddata
 
 
-def prog_plen(desired_actual):
+def prog_plen(desired_actual, amplitude=1.0):
```
jmfranck: I would say that the amplitude is not a keyword argument

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
     """
-    Takes the desired pulse length and tells the
+    Takes the desired beta (us sqrt(W)) and tells the
```
jmfranck: `ctrl-k * m` !!!

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
     user what pulse length should be programmed in order to get the actual desired
-    pulse length
+    beta
     Parameters
     ==========
     desired_actual: float
-                    the actual pulse length you wish the spincore to output,
-                    in us
+                    the actual beta you wish the spincore to output,
```
jmfranck: Can we call this parameter beta, desired beta, or just desired. Desired actual is very weird... I don't know what it means

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
     # }}}
     # {{{ prepare data into arrays for interpolation
     # gather programmed pulse lengths in array
-    plen_prog = r_[0, prog90, 2 * prog90]
+    plen_prog = r_[0, t_p]
+    sqrt_P = amplitude * np.sqrt(75)
     # assume the longest pulse is about the correct length
     # and again gather into an array
-    plen_actual = r_[0, act90, act180] * 2 * prog90[-1] / act180[-1]
+    plen_actual = r_[0, beta] * ((2 * t_p[-1] * sqrt_P) / (2 * beta[-1]))
```
jmfranck: I don't follow any of this plen stuff. I would propose that rather than doing all this calibration table stuff, you just determine the coefficients and the linear threshold in the processing code for which you currently have a PR open

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
 
     # }}}
     def zonefit(desired_actual):
-        if desired_actual < 25:
+        if desired_actual < 100:
```
jmfranck: This is what I refer to as the linear threshold. Make it into a variable, so you can set it above

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
             mask = np.ones_like(plen_prog, dtype=bool)
         else:
-            mask = plen_prog > 10
+            mask = plen_prog > 60  # where line is curved
```
jmfranck: Actually there are two parameters needed for the linear threshold. The parameter above is where we use the 10th order polynomial, and the parameter here is the regime over which we believe the data is actually linear

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,75 +3,153 @@
...
         calibration_data = nddata(plen_prog[mask], [-1], ["plen"]).setaxis(
             "plen", plen_actual[mask]
         )
         calibration_data.sort("plen")
         # fit the programmed vs actual lengths to a polynomial
-        if desired_actual < 25:
+        if desired_actual < 100:
```
jmfranck: See above

examples/90time_conv.py
```diff
@@ -3,19 +3,19 @@
...
 import matplotlib.pyplot as plt
 from numpy import r_
 import pyspecdata as psd
 from SpinCore_pp.pulse_length_conv import prog_plen
 
 # the following choice fails dramatically with the old code
-desired_plen = r_[0:50:50j]
-prog_plen = prog_plen(desired_plen)
+desired_beta = r_[10:150:50j]
+prog_plen = prog_plen(desired_beta)
```
jmfranck: This should now be run for two different amplitudes. It's probably easiest to just nest the code in a loop

examples/90time_conv.py
```diff
@@ -3,19 +3,19 @@
...
 with psd.figlist_var() as fl:
-    fl.next("Actual vs programmed plen")
-    plt.plot(prog_plen, desired_plen, "o")
-    plt.ylabel(r"actual plen / $\mu$s")
-    plt.xlabel(r"programmed plen / $\mu$s")
+    fl.next(r"Actual vs programmed $\beta$")
```
jmfranck: No, it's not the actual versus programmed beta... You want to plot the pulse length versus the desired beta

examples/90time_conv.py
```diff
@@ -3,19 +3,19 @@
...
+    plt.plot(prog_plen, desired_beta, "o")
```
jmfranck: The x and y here should be flipped, and the labels below need to be adjusted

examples/90time_conv.py
jmfranck: For clarity, delete this line here, and just call the function on the same line where you're plotting (inside the plotting function)

examples/90time_conv.py
jmfranck: As I note above, I would not define what are supposed to be the y values here as a separate variable.. Rather, here you just have your desired beta values as x and then for the y you call the function on those desired beta values

examples/calib_pulses.py
```diff
@@ -0,0 +1,91 @@
...
+with GDS_scope() as gds:
+    gds.reset()
+    gds.CH2.disp = True
+    gds.write(":CHAN1:DISP OFF")
+    gds.write(":CHAN2:DISP ON")
+    gds.write(":CHAN3:DISP OFF")
+    gds.write(":CHAN4:DISP OFF")
+    gds.CH2.voltscal = 100e-3  # set voltscale to 100 mV
+    gds.timescal(50e-6, pos=0)  # set timescale to 50 us
+    gds.write(":TIM:MOD WIND")
```
jmfranck: What does this do? Add a comment?

examples/calib_pulses.py
```diff
@@ -0,0 +1,91 @@
...
+    gds.write(":CHAN2:IMP 5.0E+1")  # set impedance to 50 ohm
+    gds.write(":TRIG:SOUR CH2")
+    gds.write(":TRIG:MOD NORMAL")  # set trigger mode to normal
+    gds.write(":TRIG:HLEV 7.5E-2")  # used in gds_for_tune which seems reasonable
+    # }}}
+    for index, val in enumerate(desired_beta):
+        t_p = spc.prog_plen(val)
```
jmfranck: I see you using this program in two ways: 

The first is to generate the dataset that you use to determine the calibration 

The second is to verify that that calibration works 

In the first case, this loop is run over a linearly spaced array of pulse lengths, and this line is not run. 

So, you can easily allow for both functionalities by changing the name of the array that you loop over above, and adding a conditional that determines what that array is set to, as well as whether or not this line is run

examples/calib_pulses.py
```diff
@@ -0,0 +1,91 @@
...
+                ("delay", config_dict["deadtime_us"]),
+            ]
+        )
+        spc.stop_ppg()
+        spc.runBoard()
+        datalist.append(gds.waveform(ch=2))
+        spc.stopBoard()
+data = psd.concat(datalist, "t_p").reorder("t")
+data.set_units("t", "s")
+data.set_prop("set_t", prog_t)
```
jmfranck: Again, if you have your conditional, you can use to determine the name of the indirect dimension as well as whether or not this property gets set

jmfranck (changes requested):

SpinCore_pp/pulse_length_conv.py
guinneak198: it doesn't make sense to do the calibration in post processing since this is needed in order to get the correct pulse length to send to SC

guinneak198: or are you saying determine the `c[::-1]` in post processing and make a table of just those values for this function?

examples/calib_pulses.py
guinneak198: This was pulled from capture_gds.py (used in generating test signal) but after reading docs we don't want this. Good catch!

SpinCore_pp/pulse_length_conv.py
jmfranck: The array of coefficients, which is what you point to, as well as the linear threshold numbers should both come from your processing script.

examples/calib_pulses.py
jmfranck: I actually am asking what it does... How do you know that you want it or not?

guinneak198: it sets the time based mode to zoom the window. We don't want to do this since I'm pretty sure it would negate the setting of the voltscale

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,78 +1,57 @@
...
-        if desired_actual < 25:
-            c = calibration_data.polyfit("plen", order=10)
-        else:
-            c = calibration_data.polyfit("plen", order=1)
-        return np.polyval(c[::-1], desired_actual)
-    ret_val = np.vectorize(zonefit)(desired_actual)
+    def zonefit(desired_beta):
+        return np.polyval(c[::-1], desired_beta)
+
+    ret_val = np.vectorize(zonefit)(desired_beta)
```
jmfranck: no -- we still  want the linear fit for very long pulses.

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        spc.stop_ppg()
+        spc.runBoard()
+        datalist.append(gds.waveform(ch=2))
+        spc.stopBoard()
+data = psd.concat(datalist, "t_p").reorder("t")
+data.set_units("t", "s")
+data.set_prop("set_t", t_p_range)
+data.set_prop("desired_betas", desired_beta)
+data.set_prop("acq_params", config_dict.asdict())
```
jmfranck: in the processing code, I use `"t_pulse"`  in place of `"t_p"`  it might be easier to just change this.  In either case, you need to label this dimension with the actual pulse lengths that you  programmed in.

Isn't this something I already said?

jmfranck (changes requested):

examples/calib_pulses.py
jmfranck: no, but I don't think we need it.

examples/90time_conv.py
```diff
@@ -3,19 +3,17 @@
...
-# the following choice fails dramatically with the old code
-desired_plen = r_[0:50:50j]
-prog_plen = prog_plen(desired_plen)
 with psd.figlist_var() as fl:
-    fl.next("Actual vs programmed plen")
-    plt.plot(prog_plen, desired_plen, "o")
-    plt.ylabel(r"actual plen / $\mu$s")
-    plt.xlabel(r"programmed plen / $\mu$s")
+    for amplitude in [(1.0), (0.5)]:
+        fl.next(r"desired $\beta$ vs pulse length")
```
jmfranck: what you are plotting is "required pulse length vs. desired β"

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+    gds.timescal(50e-6, pos=0)  # set timescale to 50 us
+    gds.write(":CHAN2:IMP 5.0E+1")  # set impedance to 50 ohm
+    gds.write(":TRIG:SOUR CH2")
+    gds.write(":TRIG:MOD NORMAL")  # set trigger mode to normal
+    gds.write(":TRIG:HLEV 7.5E-2")  # used in gds_for_tune which seems reasonable
+    # }}}
+    if calibrating:
+        t_p_range = np.linspace(0.5, 25, 25)
+    else:
+        t_p_range = spc.prog_plen(desired_beta, config_dict["amplitude"])
```
jmfranck: move this up to the top.

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+    for index, val in enumerate(t_p_range):
+        t_p = val
```
jmfranck: this is not clean code.  Just use t_p as the loop variable.

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+        prog_t.append(t_p)
```
jmfranck: not clean.  You already have an array of pulse lengths that you're looping over.

examples/calib_pulses.py
jmfranck: so, actually at the end, your conditional will come into play again.  If you are calibrating, then you  want the axis to be called t_pulse and be labeled with the pulse lengths.  If you are checking, it makes more sense to call it "beta" and label with the desired beta values.

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+)
+# }}}
+# {{{ add file saving parameters to config dict
+config_dict["type"] = "pulse_calib"
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["misc_counter"] += 1
+# }}}
+# {{{ ppg
+sqrt_P = config_dict["amplitude"] * np.sqrt(75)  # we have a 75 W amplifier
+desired_beta = np.linspace(0.5 * sqrt_P, 100, 50)
```
jmfranck: if you are calibrating, you are not calculating desired beta.  Also, desired beta should not change with the amplitude.  Which means we are not interested in this sqrt_P.

examples/calib_pulses.py
```diff
@@ -0,0 +1,98 @@
...
+tx_phases = np.r_[0.0, 90.0, 180.0, 270.0]
+Rx_scans = 1
+datalist = []
+prog_t = []
```
jmfranck: delete this `prog_t`

examples/calib_pulses.py
```diff
@@ -64,12 +66,9 @@
...
     gds.write(":TRIG:MOD NORMAL")  # set trigger mode to normal
-    if config_dict["amplitude"] > 0.5:
-        gds.CH2.voltscal = 500e-3  # set voltscale to 100 mV
-        gds.timscal(50e-6, pos=20e-6)  # set timescale to 50 us
-    else:
-        gds.CH2.voltscal = 50e-3
-        gds.timscal(100e-6, pos=250e-6)
+    # PR COMMENT: I tried to make the following so that it could be used flexibly with a range of powers
+    gds.CH2.voltscal = sqrt(nominal_power / nominal_atten * 50)
+    gds.timscal(np.max(t_pulse_range), pos=20e-6)
```
jmfranck: I tried to make the following so that it could be used flexibly with a range of powers

examples/calib_pulses.py
```diff
@@ -34,7 +34,7 @@
 )
 # }}}
 if calibrating:
-    t_pulse_range = np.linspace(
+    t_pulse = np.linspace(
```
jmfranck: this is just an array of pulse lengths, so I simplify the name

jmfranck (changes requested):

jmfranck:
I pushed more updates

examples/calib_pulses.py
guinneak198: edited to be max(t_pulse_range)*1e-6 but yes looks good

SpinCore_pp/ppg/echo.py
```diff
@@ -86,14 +89,16 @@ def run_spin_echo(
...
     """
     assert nEchoes == 1, "you must only choose nEchoes=1"
     # take the desired p90 and p180
     # (2*desired_p90) and convert to what needs to
     # be programmed in order to get the desired
     # times
-    prog_p90_us = prog_plen(p90_us)
-    prog_p180_us = prog_plen(2 * p90_us)
+    prog_p90_us = prog_plen(plen, amplitude) if plen_as_beta else plen
+    prog_p180_us = prog_plen(2 * plen, amplitude) if plen_as_beta else plen
```
jmfranck: please see what I'm doing here, and do the same for the other files in the ppg directory

examples/run_pulse_calibration.py
```diff
@@ -40,12 +41,9 @@
...
 if calibrating:
     t_pulse_us = np.linspace(
-        0.5,
-        350
-        / np.sqrt(nominal_power)
-        / config_dict[
-            "amplitude"
-        ],  # if the amplitude is small we want to go out to much longer pulse lengths
+        # if the amplitude is small we want to go out to much longer pulse lengths
+        0.5 / np.sqrt(nominal_power) / config_dict["amplitude"],
```
jmfranck, line 56: I changed this -- I assume this is OK

examples/run_pulse_calibration.py
```diff
@@ -61,7 +59,7 @@
 with GDS_scope() as gds:
     # {{{ set up settings for GDS
     gds.reset()
-    gds.autoset
+    gds.autoset # PR this is a method, and you are not calling it here
```
jmfranck: deal with comment

examples/run_pulse_calibration.py
```diff
@@ -104,6 +107,7 @@ def round_for_scope(val, multiples=1):
             config_dict["amplitude"],
             nPoints,
         )
+        # PR: is this given in ms? if so, should be reflected in variable name
```
jmfranck: deal with comment

examples/run_pulse_calibration.py
```diff
@@ -154,6 +158,7 @@ def round_for_scope(val, multiples=1):
 else:
     data.setaxis("beta", desired_beta)
     data.set_prop("programmed_t_pulse_us", t_pulse_us * 1e-6)
+# PR why have you *still* not incorporated a postproc name??
```
jmfranck: deal with comment

jmfranck (changes requested):

SpinCore_pp/ppg/generic.py
```diff
@@ -32,6 +32,7 @@ def generic(
     ret_data=None,
     amplitude=1.0,
     manual_echoes=0,
+    # PR this guy needs to be changed too!
```
jmfranck: this

SpinCore_pp/save_data.py
```diff
@@ -19,6 +19,7 @@ def save_data(dataset, my_exp_type, config_dict, counter_type, proc=True):
         config_dict pulled from the active.ini file
     counter_type: str
         type of counter you are incrementing
+    PR: need documentation for proc
```
jmfranck: this

jmfranck (changes requested):
if you take care of only the things I flag here, we should be good to go.  Please don't forget to drop a comment (in the PR is actually the easiest -- I don't mean to copy/paste onto slack unless you want to).

SpinCore_pp/ppg/generic.py
guinneak198: see [here](https://jmfrancklab.slack.com/archives/CV553TPRV/p1723756515412379)

jmfranck (changes requested):
I ran black on a bunch of files on master to simplify the diff please merge in master so that it actually is reflected here.

SpinCore_pp/ppg/echo.py
```diff
@@ -19,24 +19,16 @@
...
+    plen,
     indirect_fields=None,
     ph1_cyc=r_[0, 1, 2, 3],
     ph2_cyc=r_[0],
     ret_data=None,
-    deadtime_us=10.0,
-    deblank_us=1.0,
-    amplitude=1.0,
+    plen_as_beta=True,
 ):
```
jmfranck: if you do this, you're going to have to change *everything* that uses `run_spin_echo`, which are the following:
```
./SpinCore_pp/run_nutation_CPMG.py
./examples/run_CPMG.py
./examples/run_FS.py
./examples/run_largecyc_CPMG.py
./examples/run_var_tau.py
./examples/run_nutation.py
./examples/combined_ODNP.py
./examples/run_nutation_amplitude.py
./examples/run_Hahn_echo_mw.py
./examples/run_generic_echo.py
./examples/run_Hahn_echo.py
./examples/run_field_dep_justMw.py
./examples/run_largecyc_generic_echo.py
./Instruments/nmr_signal_gui.py
```
Is this the best way to go?

Since the generic sequence just takes raw pulse lengths, maybe it would be easier to go back to raw pulse lengths for `run_spin_echo` as well?

There's also a `run_IR` that would require a similar change.

I get why you're doing this,  but keep in mind that it will require changing a lot.

examples/demo_calibration.py
```diff
@@ -0,0 +1,27 @@
+""" Pulse length conversion
```
jmfranck, line 1: this file looks good --  no further review

examples/active.ini
```diff
@@ -1,6 +1,7 @@
 [acq_params]
 amplitude = 0.575
 p90_us = 1.7
+beta_90_s_sqrtw = 2.5e-05
```
jmfranck, line 4: this makes sense -- no further review for this file.

examples/90time_conv.py
```diff
@@ -1,21 +0,0 @@
-""" Pulse length conversion
```
jmfranck, line 1: this makes sense -- no further review for this file

SpinCore_pp/save_data.py
```diff
@@ -5,7 +5,7 @@
 import subprocess
 
 
-def save_data(dataset, my_exp_type, config_dict, counter_type):
+def save_data(dataset, my_exp_type, config_dict, counter_type, proc=True):
```
jmfranck: this makes sense -- no further review for this file

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,124 @@
...
 
-def prog_plen(desired_actual):
+def prog_plen(desired_beta, settings):
     """
-    Takes the desired pulse length and tells the
-    user what pulse length should be programmed in order to get the actual desired
-    pulse length
+    Takes the desired β (μs√(W)) and configuration file 
+    to calculate what pulse length should be programmed in order to
+    get the desired β based on the amplitude and deblank setting.
```
jmfranck, line 9: I think you need an empty line before "Parameters"

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,124 @@
...
+    else:
+        assert (
+            desired_beta[-1] < 1000e-6
+        ), "You asked for a desired beta of over 1,000 μs√W.  This is not the beta value you are looking for!!!"
+    assert (
+        settings["deblank_us"] == 50
+    ), "currently only calibrated for deblank_us = 50, so you almost definitely want to set that value in your active.ini"
+    if settings["amplitude"] == 1.0:
+        linear_threshold = 100e-6
+        c_nonlinear = [
```
jmfranck: these should all be `r_`
which means at the top, you need `from numpy import r_`

jmfranck: (`r_` gives an array -- otherwise it's a list)

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,124 @@
-from pylab import array
 import numpy as np
-from pyspecdata import r_, nddata
+import SpinCore_pp as spc
```
jmfranck: changes marked should finalize this file.

SpinCore_pp/ppg/echo.py
```diff
@@ -19,24 +19,16 @@
 
 # {{{spin echo ppg
 def run_spin_echo(
-    nScans,
+    settings,
```
jmfranck: as noted below, I think the strategy introduced here is problematic.
I think the easiest solution is to roll back to just passing the pulse length to this function.
Because that's what we do for the generic (which we expect to use more, anyways), I think that makes the most sense.

SpinCore_pp/config_parser_fn.py
```diff
@@ -199,43 +205,6 @@ class configuration(object):
             None,
             "number of IR experiments collected in the ODNP experiment",
         ),
-        # reviewed to here
```
jmfranck, line 202: I did this -- rather than adding new counters, I just did the counter thing below.  It seems to pass the config example.

jmfranck (changes requested):
this one has some issues still.

SpinCore_pp/ppg/echo.py
guinneak198: I rolled back this edit to the pulse_converting function as well to just take amplitude and deblank_us flat out rather than taking the config_dict as an argument

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,122 @@
...
-                    the actual pulse length you wish the spincore to output,
-                    in us
+    desired_beta: float
+                    the desired β you wish the spincore to output,
+                    in μs*sqrt(W)
+    amplitude: float
+                amplitude setting of the output rf pulse
+    deblank_us: float
+                deblanking period before the actual output
+                rf pulse
```
guinneak198: this was rolled back since we aren't making the config_dict an argument of echo.py it can't feed this function the config_dict so I just make these arguments flat out

examples/demo_calibration.py
guinneak198: note edit to correspond with argument of the converter being amplitude and deblank

SpinCore_pp/ppg/echo.py
```diff
@@ -97,10 +97,13 @@ def run_spin_echo(
...
-    prog_p180_us = (
-        prog_plen(2 * plen, amplitude, deblank_us) if plen_as_beta else (2 * plen)
-    )
+    if plen_as_beta:
+        settings = {"amplitude": amplitude, "deblank_us": deblank_us}
+        prog_p90_us = prog_plen(plen, settings)
+        prog_p180_us = prog_plen(2 * plen, settings)
+    else:
+        prog_p90_us = plen
+        prog_p180_us = 2 * plen
```
jmfranck: I take this approach -- which ends up being simpler than what you had done.

SpinCore_pp/pulse_length_conv.py
```diff
@@ -3,26 +3,27 @@
...
-                rf pulse
+    desired_beta : float
+        the desired β you wish the spincore to output,
+        in μs*sqrt(W)
+    settings :  dict-like
+        contains the following keys.  It's crucial that these get used in the pulse sequence.
+
+        :amplitude: float
+        :deblank_us: float
+
```
jmfranck: see [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

SpinCore_pp/pulse_length_conv.py
```diff
@@ -88,7 +89,7 @@ def prog_plen(desired_beta, amplitude, deblank_us):
             5.25471331e43,
         ]
         c_linear = r_[1.87827645e00, 1.06425500e06]
-    elif amplitude == 0.05:
+    elif settings["amplitude"] == 0.05:
```
jmfranck, line 91: rolling back to the way it was before

examples/demo_calibration.py
```diff
@@ -11,13 +11,14 @@
 from SpinCore_pp.pulse_length_conv import prog_plen
 import SpinCore_pp as spc
 
-deblank_us = 50.0  # currently the pulse length converter is calibrated only for a deblank of 50 us
+config_dict = spc.configuration("active.ini")
```
jmfranck, line 14: rolling back

examples/run_pulse_calibration.py
```diff
@@ -159,12 +157,12 @@ def round_for_scope(val, multiples=1):
...
         data[indirect, idx] = thiscapture
 if calibrating:
-    data.setaxis("t_pulse", t_pulse_us / 1e6).set_units(
+    data.setaxis("t_pulse", t_pulse_us * 1e-6).set_units(
         "t_pulse", "s"
     )  # always store in SI units unless we're wanting to change the variable name
 else:
     data.setaxis("beta", desired_beta).set_units("beta", "s√W")
-    data.set_prop("programmed_t_pulse", t_pulse_us / 1e6)  # use SI units
+    data.set_prop("programmed_t_pulse", t_pulse_us * 1e-6)  # use SI units
```
jmfranck: this is just clearer

jmfranck (changes requested):

examples/run_pulse_calibration.py
```diff
@@ -27,7 +27,7 @@
 nominal_power = 75  # in W
 nominal_atten = 1e4
 num_div_per_screen = 8
-n_lengths = 50  # number of points acquired
+n_lengths = 50  # number of pulse lengths acquired
```
jmfranck, line 33: typo

examples/run_pulse_calibration.py
```diff
@@ -42,7 +42,7 @@
 # {{{ add file saving parameters to config dict
 config_dict["type"] = "pulse_calib"
 config_dict["date"] = datetime.now().strftime("%y%m%d")
-config_dict["misc_counter"] += 1
+config_dict["pulse_calib_counter"] += 1
```
jmfranck: can be done now that I've generalized the counter variables.

jmfranck (changes requested):

jmfranck:
OK!!  There were some changes that were non-obvious and straightforward, so I just made them, and commented on them in my reviews above.

(Aside from messing up the numpy docstring -- you don't indent correctly!)

But, you need to please review my changes and make sure it's still going to work before we go ahead and merge.

jmfranck:
also, it's not passing flake -- It think a simple fix

SpinCore_pp/pulse_length_conv.py
guinneak198: went through other docstrings in this PR and edited to correspond to that outline

examples/run_pulse_calibration.py
guinneak198: save_data does this so I edited accordingly but did change name of counter as you point out

SpinCore_pp/ppg/echo.py
```diff
@@ -37,76 +37,96 @@ def run_spin_echo(
...
+        set this to None.
 
         This parameter is only used when `ret_data` is set to `None`.
-    ph1_cyc: array
-        phase steps for the first pulse
-    ph2_cyc: array
-        phase steps for the second pulse
-    ret_data: nddata (default None)
-        returned data from previous run or `None` for the first run.
+
```
jmfranck: no added space here

SpinCore_pp/config_parser_fn.py
```diff
@@ -66,6 +66,12 @@ class configuration(object):
             None,
             "90 time of the probe in microseconds.\nUsed to determine 90° 180°, etc pulses",
         ),
+        "beta_90_s_sqrtW": (
```
jmfranck, line 69: this file reviewed + OK

Instruments/nmr_signal_gui.py
```diff
@@ -157,7 +157,7 @@ def generate_data(self):
             carrierFreq_MHz=self.myconfig["carrierFreq_MHz"],
             nPoints=self.npts,
             nEchoes=self.myconfig["nEchoes"],
-            p90_us=self.myconfig["p90_us"],
```
jmfranck, line 160: this file reviewed + OK

SpinCore_pp/ppg/echo.py
```diff
@@ -37,76 +37,96 @@ def run_spin_echo(
...
+        )
         run_scans_time_list.append(time.time())
         run_scans_names.append("configure Rx")
         acq_time_ms = configureRX(
-            SW_kHz, nPoints, RX_nScans, nEchoes, nPhaseSteps
+            SW_kHz,
+            nPoints,
+            RX_nScans,
+            nEchoes,
+            nPhaseSteps,
```
jmfranck, line 127: no other comments for this file

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,122 @@
...
+    get the desired β based on the amplitude and deblank setting.
+
     Parameters
     ==========
-    desired_actual: float
-                    the actual pulse length you wish the spincore to output,
-                    in us
+    desired_beta : float
+        the desired β you wish the spincore to output,
+        in μs*sqrt(W)
```
jmfranck: is this correct?  I think it's actually s*sqrt(W)?

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,122 @@
...
+    def zonefit(desired_beta):
+        if desired_beta < linear_threshold:
+            return np.polyval(c_nonlinear[::-1], desired_beta)
         else:
-            c = calibration_data.polyfit("plen", order=1)
-        return np.polyval(c[::-1], desired_actual)
+            return np.polyval(c_linear[::-1], desired_beta)
 
-    ret_val = np.vectorize(zonefit)(desired_actual)
+    ret_val = np.vectorize(zonefit)(desired_beta)
```
jmfranck, line 137: no further comments on this file

examples/run_pulse_calibration.py
```diff
@@ -0,0 +1,172 @@
...
+If calibrating the pulse lengths, a series of pulse lengths are sent to the 
+spincore directly and the output pulse is captured on the GDS oscilloscope.
+If testing the calibration or capturing using a series of desired betas,
+the calibrating conditional should be set to false and the script will calibrate
+the pulse lengths so that the output of the amplifier produces the desired beta.
+It is very important to note that there MUST BE at least a 40 dBm
+attenuator between the RF amplifier output and the GDS oscilloscope input to
+avoid damaging the instrumentation! It is advised that the attenuator be
+calibrated using the GDS and AFG beforehand
+"""
```
jmfranck, line 17: this file requires a final review when you're sure everything works + is ready

SpinCore_pp/config_parser_fn.py
guinneak198: In testing AG realized the counter does increment properly but does not show up in the active.ini

jmfranck: That would need debugging. I would first check if you are not writing the config at the end or something like that. Because the example for the config file does show the counter increments between runs, and it needs to save to the ini file to do that.

jmfranck: Yes, this has to be a problem with the way that you're implementing it... Please see the example where I just added a counter that I call bogus counter that is entirely new and gets saved to the file

jmfranck: I just pushed this to this branch

jmfranck: I just pushed this to this branch

SpinCore_pp/config_parser_fn.py
```diff
@@ -66,6 +66,12 @@ class configuration(object):
...
+            float,
+            "acq_params",
+            None,
```
jmfranck: Give this a default

jmfranck (commented):
You left confusing comments, so waiting for a double check

SpinCore_pp/config_parser_fn.py
guinneak198: I was wrong [here](https://github.com/jmfrancklab/FLinst/pull/14#discussion_r1734648342) and we do need to increment inside this file

examples/run_pulse_calibration.py
guinneak198: deleting this was a mistake on my end and this is needed so that the counter shows up in the active ini after writing. My bad!

examples/run_largecyc_CPMG.py
```diff
@@ -22,6 +22,7 @@
 If you wish to keep the field as is without adjustment, follow
 the 'py run_CPMG.py' command with 'stayput' (e.g. 'py run_CPMG.py stayput')
 """
+
```
guinneak198, line 25: For functions there is no empty line but for example[s there is ](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)

examples/run_nutation.py
```diff
@@ -6,6 +6,7 @@
 that we are able to see when the signal rotates through 90 to 
 180 degrees.
 """
+
```
guinneak198, line 9: see above note about documentation of functions vs examples

SpinCore_pp/pulse_length_conv.py
```diff
@@ -1,79 +1,140 @@
...
+        ==========
+        desired_beta: float
+            Beta that the user wants to obtain from the output pulse.
+        
+        Returns
+        =======
+        retval: ndarray
+            Array containing the pulse length that will enable the
+            SpinCore to output the desired beta(s).
+        """
```
guinneak198, line 129: added docstring for function, maybe review and see if I messed anything up?

examples/basic_config_example.py
```diff
@@ -1,7 +1,10 @@
-"just testing/illustrating the basic function of configfiles -- really for devel purposes"
+"""Basic config dict example
+============================
+
+Just testing/illustrating the basic function of configfiles -- really
+for devel purposes
+"""
```
guinneak198, line 6: Edited docstring

guinneak198:
A lot of these are just 2 liners editing the prog_plen arguments (e.g. examples/run_CPMG.py) but I fully reviewed the run_pulse_calibration.py

guinneak198 (commented):
A lot of 2 line/ 1 line changes (hence all the files) for editing the arguments of prog_plen

examples/run_pulse_calibration.py
```diff
@@ -0,0 +1,193 @@
...
+    config_dict["SW_kHz"],
+    config_dict["acq_time_ms"],
+) = spc.get_integer_sampling_intervals(
+    config_dict["SW_kHz"], config_dict["acq_time_ms"]
+)
+# }}}
+# {{{ add file saving parameters to config dict
+config_dict["type"] = "pulse_calib"
+config_dict["date"] = datetime.now().strftime("%y%m%d")
+config_dict["pulse_calib"] += 1
```
jmfranck: This is not called "counter"... Might be my bad. 

Also, are the three lines here not handled by the file saving function? In particular, the date?

jmfranck (changes requested):
One small question/comment, then it should be good

PR #3, opened 2024-07-15
Better nutation

guinneak198:

PR #16, opened 2024-08-13
ppg to capture a single pulse on GDS scope

guinneak198:
script for very rough straightforward acquisition of a single pulse on the GDS scope


examples/simple_onepulse_capture.py
```diff
@@ -0,0 +1,96 @@
...
+    # throw some stuff out beforehand
+    thiscapture = thiscapture["t":(0, None)]["t":(None, 24e6)]
+    thiscapture *= 2
+    thiscapture["t", 0] *= 0.5
+    thiscapture.ift("t")
+    # }}}
+    data = thiscapture
+data.set_prop("programmed_t_pulse_us", t90_pulse_us)
+data.set_units("t", "s")
+data.set_prop("acq_params", config_dict.asdict())
```
jmfranck, line 124: Set postproc type

jmfranck:
[this PR](https://github.com/jmfrancklab/FLinst/pull/3) must be resolved first

jmfranck:
please see comments in code `# PR`

guinneak198:
I was under the impression we wanted this script as bare bones as possible and so had taken out the functions you mentioned giving the user control of the timescale etc without messing with function parameters. but added back in

examples/simple_onepulse_capture.py
```diff
@@ -0,0 +1,127 @@
...
+    thiscapture = gds.waveform(ch=2)
+    assert (
+        np.diff(thiscapture["t"][np.r_[0:2]]).item() < 0.5 / 24e6
+    ), "what are you trying to do, your dwell time is too long!!"
+    # {{{ just convert to analytic here, and also downsample
+    #     this is a rare case where we care more about not keeping
+    #     ridiculous quantities of garbage on disk, so we are going to
+    #     throw some stuff out beforehand
+    thiscapture.ft("t", shift=True)
+    thiscapture = thiscapture["t":(0, None)]["t":(None, 24e6)]
```
jmfranck: why isn't this just `["t":(0,24e6)]`?  If there is a reason, can you add a comment?

jmfranck (changes requested):
just one thing, and then we should be able to merge.
Thanks for getting this to such a good state!

examples/simple_onepulse_capture.py
guinneak198: was basing off of other acquisition scriptss but adjusted to (0,24e6) - I agree this way is more straightforward

PR #17, opened 2024-08-14
WK update PSD script for ch2 on scope

wfkincaid:


wfkincaid:
This is just to update the master version of collect100.py to be used with CH2 of the scope. Changes made and runs but needs to be cleaned up to pass flake

jmfranck:
I'm closing this b/c the file wasn't formatted properly and was in the wrong location to start with.

On master, I moved the file and made it so that it passes flake.  You can use that as a starting point.

My one comment was going to be that I don't understand why you changed it from asking for ODNP_NMR_comp/noise_tests to asking for just noise_tests

PR #19, opened 2024-08-20
implement version increments for RX scans

guinneak198:


examples/run_Hahn_echo.py
```diff
@@ -92,7 +92,9 @@
 )
 data.setaxis("ph1", ph1_cyc / 4)
 data.reorder(["ph1", "nScans", "t2"])
-data.set_prop("postproc_type", "spincore_SE_v1")
+data.set_prop("postproc_type", "spincore_generalproc_v1")
```
jmfranck: You can't say that the post processing type is this. This is the name of a post-processing function that is a general function that is used by many different post-processing types

examples/run_field_dep_justMw.py
```diff
@@ -65,7 +64,7 @@
 min_dBm_step = 0.5
 for x in range(len(powers)):
     dB_settings = (
-        round(10 * (np.log10(powers[x]) + 3.0) / min_dBm_step) * min_dBm_step
+        round(10 * (log10(powers[x]) + 3.0) / min_dBm_step) * min_dBm_step
```
jmfranck: Why do the import statement differently and then change this? This is just an unnecessary change.

examples/run_field_dep_justMw.py
```diff
@@ -93,7 +92,7 @@
             break
     if p.get_power_setting() < dB_settings:
         raise ValueError("After 10 tries, this power has still not settled")
-    meter_powers = np.zeros_like(dB_settings)
+    meter_powers = zeros_like(dB_settings)
```
jmfranck: This is part of the same unnecessary change referenced above

examples/run_field_dep_justMw.py
```diff
@@ -153,7 +152,7 @@
 sweep_data.set_units("t2", "s")
 
 sweep_data.name(config_dict["type"] + "_" + str(config_dict["field_counter"]))
-sweep_data.set_prop("postproc_type", "field_sweep_v3")
+sweep_data.set_prop("postproc_type", "spincore_generalproc_v1")
```
jmfranck: Again, you need to choose a different post-processing type.

jmfranck: It seems like it should just be Field sweep version 4

examples/run_largecyc_CPMG.py
```diff
@@ -191,7 +191,8 @@
 data.setaxis("nEcho", r_[0 : config_dict["nEchoes"]]).setaxis(
     "ph1", ph1 / 4
 ).setaxis("ph2", ph2 / 4).setaxis("ph_overall", ph_overall / 4)
-data.set_prop("postproc_type", "spincore_generalproc_v1")
+data.set_prop("postproc_type", "spincore_diffph_SE_v2")
```
jmfranck: Can you actually replace SE with cpmg here? I did something like this at some point, but we have had clarifying conversations about the double use of this post-processing type property.

examples/run_largecyc_generic_echo.py
```diff
@@ -136,7 +136,8 @@
     [len(ph1), len(ph_overall), -1],
 )
 data.setaxis("ph1", ph1 / 4).setaxis("ph_overall", ph_overall / 4)
-data.set_prop("postproc_type", "spincore_generalproc_v1")
+data.set_units("t2", "s")
+data.set_prop("postproc_type", "spincore_diffph_SE_v1")
```
jmfranck: Shouldn't you increment the version number here because of the n scans thing?

examples/run_var_tau.py
```diff
@@ -76,8 +77,8 @@
     ret_data=None,
 )
 mytau_axis = var_tau_data.getaxis("indirect")
-mytau_axis[0]["tau_adjust"] = tau_adjust
-mytaus_axis[0]["tau"] = tau
+mytau_axis[0]["tau_adjust"] = tau_adjust_range[0]
+mytau_axis[0]["tau"] = tau[0]
```
jmfranck, line 81: What is going on here? This is a non-rhetorical question to be replied to in the pr

jmfranck (changes requested):

jmfranck:
I requested several changes but once those are taken care of we should be good to merge

examples/run_var_tau.py
guinneak198: if you compare the red line 79 to 80 you see there is a typo, these changes just address. Also tau_adjust_range and tau are lists and so we want to place the first index of the list into the first index of the indirect - not the whole list

examples/run_field_dep_justMw.py
guinneak198: I compared to the load_data in the master branch and there is not a field_sweep_v3 yet so this change is actually not necessary!

examples/run_largecyc_generic_echo.py
guinneak198: again checked with load_data on master - v2 and v3 already exist and are not appropriate for this data

PR #23, opened 2024-08-30
add deblank to echo args in gui

guinneak198:


jmfranck (commented):
This looks like it was run with and incorrect black configuration file. It is re-wrapping the lines so that they are longer than before

jmfranck:
Actually, I need to check that. Give me an hour

PR #24, opened 2024-09-19
implement plen argument

guinneak198:
now that we changed to feeding beta instead of p90 to echo and IR I adjusted the arguments fed for all scripts used in odnp


examples/combined_ODNP.py
```diff
@@ -99,19 +111,25 @@
...
+    could try reducing the acq_time_ms to %f"
     % total_points,
     config_dict["acq_time_ms"] * 16384 / total_points,
 )
 total_pts = len(IR_ph2_cyc) * len(IR_ph1_cyc) * nPoints
 assert total_pts < 2**14, (
-    "For IR: You are trying to acquire %d points (too many points) -- either change SW or acq time so nPoints x nPhaseSteps is less than 16384\nyou could try reducing the acq_time_ms to %f"
+    "For IR: You are trying to acquire %d points (too many points) -- either\
+    change SW or acq time so nPoints x nPhaseSteps is less than 16384\nyou\
+    could try reducing the acq_time_ms to %f"
```
jmfranck: If you try printing this in ipython (always a good idea), you will see that it's a bit weird.  I would recommend:
```
mystring = ("this "
                     "does not have "
                     "weird spaces. "
                     "But, pay attention "
                     "to how I added "
                     "the space at the end "
                     "where needed.")
```

examples/active.ini
```diff
@@ -1,10 +1,10 @@
 [acq_params]
-amplitude = 0.575
```
guinneak198, line 2: picked an amplitude that is calibrated

examples/active.ini
```diff
@@ -1,10 +1,10 @@
...
+amplitude = 0.2
 p90_us = 1.7
 beta_90_s_sqrtw = 2.5e-05
 deadtime_us = 10.0
 tau_us = 3500.0
-deblank_us = 1.0
```
guinneak198, line 7: calibration needs deblank = 50

examples/active.ini
```diff
@@ -13,7 +13,7 @@ carrierfreq_mhz = 14.895116358000001
 nechoes = 1
 startconstant = 0.15
 stopconstant = 2.0
-gamma_eff_mhz_g = 0.0042419157275600345
+gamma_eff_mhz_g = 0.004287070716629787
```
guinneak198, line 16: this is more up to date with most recent ratio

PR #26, opened 2024-10-09
fix coherence pathway used for FIRs, update run_field_dep indirect

wfkincaid:
improve/update acq scripts for running a DNP

### DIFF
--- a/examples/run_field_dep_justMw.py
+++ b/examples/run_field_dep_justMw.py
@@ -122,7 +122,7 @@ with power_control() as p:
         )
         myfreqs_fields = sweep_data.getaxis("indirect")
         myfreqs_fields[0]["Field"] = first_B0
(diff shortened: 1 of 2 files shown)


examples/run_field_dep_justMw.py
```diff
@@ -122,7 +122,7 @@
         )
         myfreqs_fields = sweep_data.getaxis("indirect")
         myfreqs_fields[0]["Field"] = first_B0
-        myfreqs_fields[0]["carrierFreq"] = config_dict["carrierFreq_MHz"]
+        myfreqs_fields[0]["carrierFreq"] = first_B0 * config_dict["gamma_eff_MHz_G"] # looking at what is done above, I should just be able to set this to `carrierFreq_MHz`
```
jmfranck: for clarity and conciseness, just use `carrierFreq_MHz` from above

PR #27, opened 2024-10-23
typo where node was saved as signal

guinneak198:


examples/SC_noise_capture.py
```diff
@@ -116,7 +116,7 @@ def collect(config_dict):
                 .setaxis("t", time_axis)
                 .set_units("t", "s")
                 .setaxis("nScans", np.zeros(config_dict["nScans"]))
-                .name("signal")
+                .name("noise")
```
jmfranck: can we call it `noisecapture` or something like that? calling it just noise is weird to me

jmfranck (changes requested):

PR #28, opened 2024-10-28
Capture gds

guinneak198:
script for calibrating test signal


examples/calibrate_test_signal.py
```diff
@@ -0,0 +1,122 @@
...
+    print(
+        "Make sure the input_Vamp here is accurate so I can calculate the"
+        " right voltage ratio!"
+    )
+    input_Vamp = 500e-3  # this should be calculated prior to calibrating
+    #                      the attenuator assemblies
+    assert input_Vamp < 505e-3, (
+        "That's way too high of an input peak voltage! Make sure it's in units"
+        " of V and not mV!"
+    )
```
guinneak198, line 43: gives a little leeway incase the actual is slightly larger than 1 V (GDS can handle 1.25 V)

examples/calibrate_test_signal.py
```diff
@@ -0,0 +1,122 @@
...
+else:
+    input_Vamp = expected_Vamp
```
guinneak198, line 45: implies no attenuator/loss of signal so expected = input

examples/calibrate_test_signal.py
```diff
@@ -0,0 +1,116 @@
...
+            % (Vamp / 1e-3, input_Vamp / Vamp),
+            transform=gca().transAxes,
+        )
+    else:
+        text(
+            0.5,
+            0.02,
+            s=r"$V_{amp} = %0.6f$ mV" % (Vamp / 1e-3),
+            transform=gca().transAxes,
+        )
```
guinneak198, line 116: voltage ratio not needed if no attenuators are attached

examples/calibrate_test_signal.py
```diff
@@ -0,0 +1,116 @@
...
+
+1. Calculating the raw output of the AFG (no attenuator assemblies). For this
+step, simply set the calibrating_atten to False. Note that for the next step
+you need to set the input_Vamp to the value calculated in this step.
+
+2. Calculating the voltage ratio of the attenuator assemblies (placed between
+the AFG output and the GDS). Prior to running the script the user needs to
+have a rough estimate of the expected_Vamp when the attenuator assemblies are
+in place. The calculated voltage ratio and dB are printed on the final
+generated plot.
```
guinneak198, line 19: not sure if steps are okay in docstring (couldn't find much on google). Feel free to edit but the gist is that first you calculate AFG --> GDS to get the actual V_input (calibrating_atten=False) and then that value is used in calculating the actual voltage ratio (V_input / V_atten)

guinneak198: Thinking about this I think this means that the cells in Verify.py showing the AFG-->GDS is unnecessary since it's included in this script.

jmfranck:
you do not "calibrate" test signal.  you simply capture the signal on the oscilloscope, and measure its analytic amplitude, so that language is problematic. Here is the capture_GDS.py that I had written in my notebook.  It seems that you've deleted a lot of the configuration stuff and it's not clear why.   It also unclear how the extra code for "calibrating attenuators" vs. just measuring voltage is helpful.  You measure the voltage without the attenuators, and then the voltage with the attenuators, and then you have the attenuation.  Please also see comments in verify.py

```
from Instruments import *
from pyspecdata import *
from serial.tools.list_ports import comports
from pylab import axhline, text, gca
import numpy as np

expected_amp = 7e-3
input_amp = 998e-3
with figlist_var() as fl:
    with GDS_scope() as g:
        # {{{ choose settings we want
        g.reset()
        g.CH1.disp = True
        g.CH2.disp = True
        g.write(":CHAN1:DISP ON")
        g.write(":CHAN2:DISP OFF")
        g.write(":CHAN3:DISP OFF")
        g.write(":CHAN4:DISP OFF")
        g.CH1.voltscal = (
            expected_amp * 1.1 / 5
        )  # set to a little more than amp/5
        g.timscal(50e-9, pos=0)
        g.write(":TIM:MOD WIND")
        print("window scale:")
        print(g.respond(":TIM:WIND:SCAL?"))
        g.write(":CHAN1:IMP 5.0E+1")
        g.write(":TRIG:SOUR CH1")
        g.write(":TRIG:MOD AUTO")  # or normal
        # g.write(':TRIG:HLEV 7.5E-2')
        g.write(":ACQ:MOD HIR")
        g.write(":CURS:MOD HV")
        g.write(":CURS:SOUR CH1")
        print("cursor mode:", g.respond(":CURS:MOD?"))
        mycursors = (expected_amp, -expected_amp)
        g.write(":CURS:V1P " + ("%0.2e" % mycursors[0]).replace("e", "E"))
        g.write(":CURS:V2P " + ("%0.2e" % mycursors[1]).replace("e", "E"))
        print("cursor:", g.respond(":CURS:V1P?"))
        print("cursor:", g.respond(":CURS:V2P?"))
        # }}}
        datalist = []
        for j in range(1, 2):  # this dumb but leave to keep flexible
            print("trying to grab data from channel", j)
            g.write(":SING")
            datalist.append(g.waveform(ch=j))
        data = concat(datalist, "ch").reorder("t")
        data.set_units("t", "s")
    fl.next("data from all channels, raw")
    fl.plot(data, label="orig")
    N = ndshape(data)["t"]
    # {{{ convert to analytic signal
    data.ft("t", shift=True)
    data = data["t":(0, None)]
    data *= 2
    data["t", 0] *= 0.5
    data.ift("t")
    # }}}
    # {{{ show the manual cursor positions
    for y in mycursors:
        axhline(y=y, color="k", alpha=0.5)
    # }}}
    fl.plot(abs(data), label="analytic")
    ph = (data["t", 1:] / data["t", :-1]).angle.sum("t").item()
    Dt = np.diff(data.getaxis("t")[r_[0, -1]]).item()
    frq = ph / Dt / 2 / pi
    # {{{ now, filter the signal
    data.ft("t")
    data["t" : (None, frq - 5e6)] = 0
    data["t" : (frq + 5e6, None)] = 0
    data.ift("t")
    # }}}
    fl.plot(data, label="filtered analytic signal")
    fl.plot(abs(data), label="filtered analytic signal")
    Vamp = abs(data["t":(1e-6, 4e-6)]).mean("t").real.item()
    text(
        0.5,
        0.75,
        s="$V_{amp}=%0.4f$ mV, ratio %0.6g, dB %0.6f"
        % (Vamp / 1e-3, input_amp / Vamp, 20 * np.log10(input_amp / Vamp)),
        transform=gca().transAxes,
    )
```

PR #31, opened 2025-01-30
Capture rec resp

guinneak198:
scripts for capturing datasets needed for characterizing receiver response (AFG to GDS and AFG to SC)


examples/AFG_output_to_GDS.py
```diff
@@ -0,0 +1,73 @@
+from Instruments import AFG, GDS_scope
```
jmfranck, line 10: these need docstrings

jmfranck (changes requested):
Everything in examples always needs a doc string

guinneak198:
so issues with wifi during travel I think caused me to ask for a review before the changes were pushed but anyway it's all there now (with docstrings!)

jmfranck (changes requested):
I realize these might not be exactly the same as the ones in the paper, but I want to diff them. Which should I diff?

guinneak198:
`$ git diff ../noise_density/MWEs/afg_vary_frq.py examples/AFG_output_to_SC.py` and `$ git diff ../noise_density/MWEs/GDS_var_freq.py examples/AFG_output_to_GDS.py`

PR #33, opened 2025-02-05
rather than using dip lock, just use the frequency from the mw tuning…

jmfranck:
… GUI


.github/workflows/flake8.yml
```diff
@@ -28,10 +28,10 @@ jobs:
           git fetch origin master
           if [ "${{ github.event_name }}" == "pull_request" ]; then
             git fetch origin ${{ github.event.pull_request.head.ref }}
+          else
+            git fetch origin ${{ github.event.before }}
           fi
```
wfkincaid, line 35: used in shell scripting to end an `if` block (didn't know this before)

.github/workflows/flake8.yml
```diff
@@ -28,10 +28,10 @@ jobs:
           git fetch origin master
           if [ "${{ github.event_name }}" == "pull_request" ]; then
             git fetch origin ${{ github.event.pull_request.head.ref }}
+          else
+            git fetch origin ${{ github.event.before }}
```
wfkincaid: `if` checks if there new commits/changes and pulls, `else` just fetchs if branch is already up-to-date?

wfkincaid: I think the `if` block on lines 29-30 merges master into another branch if you are on a different branch than master to help bring the current directory up-to-date.

.github/workflows/flake8.yml
```diff
@@ -40,12 +40,12 @@ jobs:
             MYCMD="git diff --name-only origin/${{ github.base_ref }}...origin/${{ github.event.pull_request.head.ref }} --diff-filter=AM" 
             echo "$MYCMD"
             echo "$($MYCMD)"
-            CHANGED_FILES=$($MYCMD | grep '\.py$' | tr '\n' ' ')
+            CHANGED_FILES=$($MYCMD | grep '\.py$' | sed '/__init/d' | tr '\n' ' ')
```
wfkincaid, line 48: I think this searches for or searches and replaces `__init`, `grep '\.py$' searches for any python files too?

Instruments/bridge12.py
```diff
@@ -21,7 +21,9 @@ def __init__(self, *args, **kwargs):
         # Grab the port labeled as Arduino (since the Bridge12 microcontroller is an Arduino)
         cport = comports()
         if type(cport) is list and hasattr(cport[0], "device"):
-            portlist = [j.device for j in comports() if "Arduino Due" in j.description]
+            portlist = [
+                j.device for j in comports() if "Arduino Due" in j.description
+            ]
```
wfkincaid, line 29: flake (?)

Instruments/bridge12.py
```diff
@@ -46,7 +48,9 @@ def bridge12_wait(self):
         # time.sleep(5)
         def look_for(this_str):
             for j in range(1000):
-                a = self.read_until((this_str + "\r\n").encode("utf-8")).decode("utf-8")
+                a = self.read_until(
+                    (this_str + "\r\n").encode("utf-8")
+                ).decode("utf-8")
```
wfkincaid, line 61: flake

Instruments/bridge12.py
```diff
@@ -273,7 +277,8 @@ def rxpowerdbm_float(self):
         """read the integer value for the Rx power -- loops three times to
         check a consistent Rx is being read.
 
-        If all three tries are above safe_rx_level_int, triggers a safety interlock."""
+        If all three tries are above safe_rx_level_int, triggers a safety interlock.
+        """
```
wfkincaid: flake

Instruments/bridge12.py
```diff
@@ -347,7 +352,8 @@ def set_freq(self, Hz):
                     return
             raise RuntimeError(
                 "After checking status 10 times, I can't get the "
-                "frequency to change -- result is %d setting is %d" % (result, setting)
+                "frequency to change -- result is %d setting is %d"
+                % (result, setting)
```
wfkincaid, line 378: flake

Instruments/bridge12.py
```diff
@@ -399,7 +405,9 @@ def robust_int_response(self, cmd, numtries=10):
                     print(
                         "WARNING (robust try",
                         j,
-                        '): B12 is spewing garbage!!: "' + retval.decode() + '"',
+                        '): B12 is spewing garbage!!: "'
+                        + retval.decode()
+                        + '"',
```
wfkincaid, line 432: flake

Instruments/bridge12.py
```diff
@@ -409,7 +417,9 @@ def robust_int_response(self, cmd, numtries=10):
                 #                           NOTE: I can check for stuff by looking at in_waiting,
                 #                                 but I need to remember that it takes time after a write
                 #                                 command for stuff to move into the buffer
-        raise ValueError("I tried running 10 times and couldn't get an integer!!!")
+        raise ValueError(
+            "I tried running 10 times and couldn't get an integer!!!"
+        )
```
wfkincaid, line 449: flake

Instruments/bridge12.py
```diff
@@ -432,7 +442,9 @@ def freq_sweep(self, freq, dummy_readings=1, fast_run=False):
         txvalues: array
             An array of floats, same length as freq, containing the transmitted power in dBm at each frequency.
         """
-        self.write(b"screen 2\r")  # change to the screen that shows the reflection
+        self.write(
+            b"screen 2\r"
+        )  # change to the screen that shows the reflection
```
wfkincaid, line 476: flake

Instruments/bridge12.py
```diff
@@ -499,13 +511,17 @@ def lock_on_dip(
                 self.set_amp(True)
                 freq = r_[ini_range[0] : ini_range[1] : ini_step]
                 logger.info(
-                    "ini range: " + str(ini_range) + "ini step: " + str(ini_step)
+                    "ini range: "
+                    + str(ini_range)
+                    + "ini step: "
+                    + str(ini_step)
```
wfkincaid, line 548: flake

Instruments/bridge12.py
```diff
@@ -499,13 +511,17 @@ def lock_on_dip(
...
                 )
                 logger.info("Did not find previous 10 dBm run, running now")
                 self.set_power(10.0)
                 rx, tx = self.freq_sweep(freq)
             rx_dBm, freq = [
-                self.tuning_curve_data["%gdBm_%s" % (10.0, j)] for j in ["rx", "freq"]
+                self.tuning_curve_data["%gdBm_%s" % (10.0, j)]
+                for j in ["rx", "freq"]
```
wfkincaid, line 555: flake

Instruments/bridge12.py
```diff
@@ -670,7 +704,8 @@ def soft_shutdown(self):
         self.set_amp(False)
         self.set_wg(False)
         self.frq_sweep_10dBm_has_been_run = False
-        del self.freq_bounds
+        if hasattr(self, "freq_bounds"):
+            del self.freq_bounds
```
wfkincaid, line 758: Is this a catch in that sometimes it would fail if frequency bounds weren't set?

Instruments/power_control.py
```diff
@@ -96,6 +96,18 @@ def send(self, msg):
...
+        To use this, you must be at 10 dBm or less!
+
+        Parameters
+        ==========
+        freq : float
+            Frequency in Hz
+        """
+        self.send("SET_FREQ %f" % freq)
+        return
+
```
wfkincaid, line 110: new function to hard set the B12's frequency for an experiment instead of always relying on dip lock. Enables us to stay at a frequency that we have tuned to instead of B12 re-determining a dip each time

Instruments/power_control_server.py
```diff
@@ -85,6 +85,14 @@ def process_cmd(cmd, this_logobj):
...
                             nsecs += time.time()
                             print("took", j, "tries and", nsecs, "seconds")
+                        if args[0] == b"SET_FREQ":
+                            current_power = b.power_float()
+                            if current_power > 10:
+                                raise ValueError(
+                                    "to manually set the power, you must be at"
+                                    " 10 dBm or less!"
+                                )
+                            b.set_freq(float(args[1]))
```
wfkincaid, line 137: Error detecting for using the new function above to manually set B12 frequency. Serves as a check to make sure the current power is not too high before trying to hard set a frequency value

examples/combined_ODNP.py
```diff
@@ -103,10 +105,10 @@
...
 )
 T1_node_names = ["FIR_%ddBm" % j for j in T1_powers_dB]
-logger.info("dB_settings", dB_settings)
-logger.info("correspond to powers in Watts", 10 ** (dB_settings / 10.0 - 3))
-logger.info("T1_powers_dB", T1_powers_dB)
-logger.info("correspond to powers in Watts", 10 ** (T1_powers_dB / 10.0 - 3))
+print("dB_settings", dB_settings)
+print("correspond to powers in Watts", 10 ** (dB_settings / 10.0 - 3))
+print("T1_powers_dB", T1_powers_dB)
+print("correspond to powers in Watts", 10 ** (T1_powers_dB / 10.0 - 3))
```
wfkincaid: print these values instead of sending straight to log

wfkincaid: while saving as info to the log may have been useful for keeping track of the script progress and the program's current state, not necessarily needed. Printing is also very useful in the moment for someone to see while running at the NMR computer

examples/combined_ODNP.py
```diff
@@ -208,8 +210,10 @@
 #   the time axis, or smaller than the first)
 ini_time = time.time()
 vd_data = None
+logger.debug('starting T1s')
```
wfkincaid: Instead of recording more excessive amounts of info to the log, record just instance of the program's state as markers if needed for debugging

examples/combined_ODNP.py
```diff
@@ -208,8 +210,10 @@
...
 for vd_idx, vd in enumerate(vd_list_us):
     # call A to run_IR
+    logger.debug(f'T1 #{vd_idx}')
```
wfkincaid: same as line 213

examples/combined_ODNP.py
```diff
@@ -266,7 +270,7 @@
 # file itself
 vd_data.hdf5_write(filename, directory=target_directory)
 # }}}
-logger.debug("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
+logger.info("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
```
wfkincaid: `info` preferred over `debug`, want to verify we are using the correct information (file gets saved, in the desired directory). `debug` more useful for identifying problems

examples/combined_ODNP.py
```diff
@@ -275,13 +279,9 @@
...
-    # JF points out it should be possible to save time by removing this (b/c we
-    # shut off microwave right away), but AG notes that doing so causes an
-    # error.  Therefore, debug the root cause of the error and remove it!
-    retval_thermal = p.dip_lock(
-        config_dict["uw_dip_center_GHz"] - config_dict["uw_dip_width_GHz"] / 2,
-        config_dict["uw_dip_center_GHz"] + config_dict["uw_dip_width_GHz"] / 2,
-    )
+    # we do not dip lock or anything here, because we assume
+    # uw_dip_center_GHz stores the frequency of the center of the cavity
+    # resonance, which was set from the microwave tuning gui
```
wfkincaid, line 285: last comment here pointed to an error caused by removing the `dip_lock` process, this has now been addressed?

wfkincaid: Otherwise, I think this makes sense; use the set frequency instead of rerunning `dip_lock`, especially right before turning off microwaves

examples/combined_ODNP.py
```diff
@@ -290,7 +290,9 @@
     # Run the actual thermal where the power log is recording. This will be
     # your thermal for enhancement and can be compared to previous thermals if
     # issues arise
+    logger.debug('about to start thermal')
```
wfkincaid: add log to track when thermals start, useful if error does arise (as alluded to) with not running `dip_lock`

examples/combined_ODNP.py
```diff
@@ -290,7 +290,9 @@
...
     for j in range(thermal_scans):
+        logger.debug(f'thermal {j}')
```
wfkincaid: log thermal scans

examples/combined_ODNP.py
```diff
@@ -15,11 +15,13 @@
     progressive power saturation dataset, and a log of the power over time
     saved as nodes in an h5 file.
 """
+
 from numpy import r_, zeros_like
 from pyspecdata.file_saving.hdf_save_dict_to_group import (
     hdf_save_dict_to_group,
 )
 import pyspecdata as psd
+from pyspecdata import strm
```
wfkincaid, line 24: `strm` maps(?) and joins things (ints, floats, strings, etc) as a single string with arguments separated by spaces

examples/combined_ODNP.py
```diff
@@ -319,21 +321,23 @@
...
-            "SETTING THIS POWER",
+        logger.debug(strm(
+            "setting this power for E(p)",
             this_dB,
             "(",
             dB_settings[j - 1],
             powers[j],
             "W)",
-        )
+        ))
```
wfkincaid: join all these together (were already) as one large string more organized/legible in the log

examples/combined_ODNP.py
```diff
@@ -319,21 +321,23 @@
...
-                + config_dict["uw_dip_width_GHz"] / 2,
-            )
+            # Again, no dip lock because we assume the microwave tuning
+            # GUI handled finding the cavity frequency.
+            #
+            # This is not only faster, but it ensures that the
+            # uw_dip_center_GHz stores the ACTUAL B12 frequency that we
+            # use
+            p.set_power(0)  # set to 0 dBm
+            p.set_freq(config_dict["uw_dip_center_GHz"] * 1e9)
```
wfkincaid, line 343: again, we want to use the determined B12 frequency and not rerun a `dip_lock`, explanation for this included

examples/combined_ODNP.py
```diff
@@ -404,17 +408,12 @@
...
-        # {{{ make small steps in power if needed
-        if this_dB - last_dB_setting > 3:
-            smallstep_dB = last_dB_setting + 2
-            while smallstep_dB + 2 < this_dB:
-                p.set_power(smallstep_dB)
-                smallstep_dB += 2
+        logger.debug(strm(
+            "setting this power for T1(p)",
+            this_dB,
+        ))
```
wfkincaid: Why take out the small power steps, is this no longer needed? Was the small power steps for the old B12?

wfkincaid: also create a log of of what power is actually being set during the collection of T1s

examples/get_ready.sh
```diff
@@ -11,11 +11,13 @@ cd /c/apps-su/FLinst
 echo -e "I'm in this branch of FLinst:\n$(git rev-parse --abbrev-ref @)"
 echo -e "is it clean?:\n$(git status -s)"
 git pull
+cp examples/get_ready.sh $orig_dir
```
wfkincaid, line 14: import a new version of itself (if/when it is updated)

examples/get_ready.sh
```diff
@@ -11,11 +11,13 @@ cd /c/apps-su/FLinst
...
 cp examples/gds_for_tune.py $orig_dir
 cp examples/collect_SC.py $orig_dir
 cp examples/run_generic_echo.py $orig_dir
 cp examples/run_nutation.py $orig_dir
 cp examples/run_field_dep_justMW.py $orig_dir
+cp examples/combined_ODNP.py $orig_dir
```
wfkincaid, line 20: include changes made to how ODNP experiments are run

pyproject.toml
```diff
@@ -24,5 +24,10 @@ power_control_server = "Instruments.power_control_server:main"
 quit_power_control = "Instruments.just_quit:main"
 FLInst = "Instruments.cmd:cmd"
 
+[tool.black]
+line-length = 79
+target-version = ['py36', 'py37', 'py38', 'py39']
+preview = true
+
```
wfkincaid, line 31: this is a config file for this branch's (FLInst) environment (?), add section to roughly apply black for automated cleaning up of scripts

.github/workflows/flake8.yml
jmfranck: This mostly had to do with which variables GitHub defines (and does not) when it checks the pull request (the whole branch) vs the individual commit

jmfranck: This mostly had to do with which variables GitHub defines (and does not) when it checks the pull request (the whole branch) vs the individual commit

.github/workflows/flake8.yml
jmfranck: It's saying don't do flake on __init__.py

Instruments/bridge12.py
jmfranck: Yes. Specifically, if we didn't run dip lock

Instruments/power_control.py
jmfranck: Exactly

Instruments/power_control_server.py
jmfranck: Yes specifically, power needs to be 10 or lower if you try to change the frequency

examples/combined_ODNP.py
jmfranck: Yes, in general we try to use log and avoid print, but I realized this is an exception since we are actually asking the user to confirm

examples/combined_ODNP.py
jmfranck: Info is printed out. Debug only goes to the file (if we've run `init_logging`)

examples/combined_ODNP.py
jmfranck: AG's comment is too vague to be addressed in detail I think

examples/combined_ODNP.py
jmfranck: Please run black... I'm realizing there are single quotes in this newer stuff

examples/combined_ODNP.py
jmfranck: The business of breaking up the steps is now handled by the server (was before this PR). So this was not needed here

pyproject.toml
jmfranck: I don't know if it works that way in Windows, but I'm Linux, this sets the default options for black for the project.

.github/workflows/flake8.yml
wfkincaid: does `sed` exclude `__init__.py` from the list of changed files?

wfkincaid:
I think my only question that remains is what `sed` means in `sed '/__init/d'` (I thought it meant "find and remove" but your explanation seems more like "find an ignore"). I just want to understand this in the context of `CHANGED_FILES`

PR #38, opened 2025-06-13
PS Code

atahangarip:


Instruments/HallProbe_ProLogix_new.py
```diff
@@ -0,0 +1,103 @@
...
+    def gpib_write(self, cmd):
+        """Write a GPIB command without expecting a return value."""
+        return self.send(cmd)
+
+    def gpib_query(self, cmd):
+        """Write a GPIB query and return its result."""
+        return self.send(cmd)
+
+
+class LakeShore475:
```
jmfranck: remember -- do not do this, where you commit two files that correspond to two versions.  Git stores the entire history, so if you think a new version is better, then you just *replace* the existing file.  Making a file called `_new` or `version2` etc are typically signs that you are not using git correctly.

Instruments/HallProbe_ProLogix_new.py
```diff
@@ -1,56 +1,6 @@
 import socket
 import time
 
-class PrologixEthernet:
```
jmfranck: I delete this because it's attempting to duplicate the functionality of `gpib_eth` which already exists and already works perfectly.

Instruments/HallProbe_ProLogix.py
```diff
@@ -0,0 +1,101 @@
+import time
+
+class HallProbeController:
```
jmfranck: This is never going to work, because it doesn't inherit from `gpib_eth`.  Remember, use these files as examples, where I show the file name, followed by the class definition which shows the inheritance:

gigatronics.py: `class gigatronics (gpib_eth):`

HP8672A.py: `class HP8672A (gpib_eth):`

HP6623A.py: `class HP6623A (gpib_eth):`

**vocabulary word**: you need to be able to define "inheritance" in the context of object-oriented programming in order to understand what I'm saying here.  You definitely want to look that up, and get some examples of how it works.

Instruments/HallProbe_ProLogix_new.py
jmfranck: so, here, after looking at my comments, you're going to decide on which one is more likely to work, call that one "HallProbe_ProLogix.py" and then delete the other one.

jmfranck:
btw, never mind that it fails the flake linting checks for now (you said you received these by email).  Those are formatting checks that we worry about when we're actually trying to merge the code back into the master/main branch.

Instruments/HallProbe_ProLogix.py
```diff
@@ -0,0 +1,101 @@
...
+        self.sensitivity = None  # mV/kG, filled after programming
+
+    def _write(self, cmd: str):
+        """Helper to send Prologix commands."""
+        self.gpib.write(f"{cmd}\n") #A write command as you suggested
+    
+    def _respond(self, cmd: str) -> str: #Takes the responses with a write command. Returns the stripped response.
+        """Helper to send a query and return the response."""
+        self._write(cmd)
+        return self.gpib.read().strip()
```
jmfranck: these appear to be routines that are designed to ease communication with the prologix, and are not following along with the examples in the modules that I note [above](https://github.com/jmfrancklab/FLinst/pull/38/files#r2159330545).

Instruments/HallProbe_ProLogix.py
```diff
@@ -0,0 +1,101 @@
...
+        Follows the Lake Shore manual §2.1 sequence.
+        """
+        
+        # Enter user-cable programming mode
+        self._write("++mode 0")  # Put ProLogix into direct-pass-through
+        self._write("++savecfg 0")  # Don't save the ProLogix config yet
+        self._write("++addr 0")  # Cable programming livese at GPIB address 0
+
+        # Start MCBL(user-cable) programming
+        self._write("++help")  # ensure controller is in CLI (optional -says ChatGPT)
```
jmfranck: again, this is stuff that is handled **automatically** by `gpib_eth.py`.

jmfranck: I should explain further, since there are many errors of this sort throughout (too many to mark individually):

All of these `++` commands are directives **to the prologix device** (as opposed to the hall probe).  They control the format of the communication with the GPIB network.

When following the [examples noted above](https://github.com/jmfrancklab/FLinst/pull/38/files#r2159330545) you should **never** issue manual `++` commands, since the `gpib_eth` class handles them automatically.

You *do* need to tell the prologix class some info so that it knows what exact `++` commands to issue.  For example, here (from `gigatronics.py`):
```
    def __init__(self,prologix_instance=None,address=13):
        super().__init__(prologix_instance,address)
```
we are telling it that the default GPIB address is 13.   And we call the `gpib_eth` initialization routine with that address (either 13 if we just use the default, or the address that we pass to the `gigatronics` class, if we pass an `address=xx` argument.  (Note that because we *inherit* from `gpib_eth`, `super().__init__(` calls the initialization routine from `gpib_eth`)

Instruments/HallProbe_ProLogix.py
jmfranck: note this same comment applies to the class that's defined in the other file.

jmfranck:
the TL;DR here is that you are learning about *inheritance* the hard way (but that's probably the good way to learn it).  The code for your Hall probe class should actually be pretty simple with relatively few lines of code, because your new class should inherit from `gpib_eth`, which handles all of the difficult details of communicating with the prologix device, leaving only the need to pull the relevant GPIB commands from the manual.

Instruments/HallProbe_ProLogix.py
jmfranck: (all that kind of stuff is handled by `gpib_eth`, which you should be inheriting from)

Instruments/HallProbe_ProLogix.py
jmfranck: note I edited the first line of the preceding comment.

jmfranck: (actually I edited the whole thing)

PR #39, opened 2025-11-13
AG nmr gui

atahangarip:


Instruments/gds.py
```diff
@@ -265,10 +265,9 @@ def upto_hashtag():
         logger.debug(strm("acquisition parameters", param))
         data_array *= float(param.pop("Vertical Scale"))
         data_array *= 5 * 1.032  # this is empirical
-        if not all(np.isfinite(x_axis)):
+        if not np.all(np.isfinite(x_axis)):
```
jmfranck: `all` should be builtin, not numpy

Instruments/gds.py
```diff
@@ -265,10 +265,9 @@ def upto_hashtag():
...
             raise ValueError(
-                "your x axis is not finite!! len(data_array) is %s and"
-                " 'Sampling Period' is %s"
+                "your x axis is not finite!! len(data_array) is %s and 'Sampling Period' is %s"
```
jmfranck: this was probably done by black or something -- I would just convert to an f-string

Instruments/genesys.py
```diff
@@ -310,7 +310,9 @@ def I_meas(self):
         - **Assignment**: Not supported.
         - **Deletion**: Not supported.
         """
-        return float(self.respond("MEAS:CURR?"))
+        retval = self.respond("MEAS:CURR?")
+        print("Current measured: ", retval)
```
jmfranck: don't include this in the PR

Instruments/nmr_signal_gui.py
```diff
@@ -41,11 +41,17 @@
 import pyspecdata as psp
 import matplotlib.backends.backend_qt5agg as mplqt5
 from matplotlib.figure import Figure
+from Instruments import (
+    genesys,
+    LakeShore475,
+    prologix_connection,
+    power_control,
```
jmfranck: I don't think that we want `power_control` for this module.

Instruments/nmr_signal_gui.py
```diff
@@ -120,26 +126,67 @@ def on_pick(self, event):
...
-            and abs(Field - self.prev_field) < 50.0 / gammabar_H * 1e4
-        ):
-            print("You seem to be within 50 Hz, so I'm not changing the field")
-        else:
-            Field = self.xepr.set_coarse_field(Field)
-            self.prev_field = Field
-            print("about to return from set_field_conditional")
+        config_dict = SpinCore_pp.configuration("active.ini")
+        I_setting = Field * config_dict["current_v_field_A_G"]
+        with genesys("192.168.0.199") as g:
```
jmfranck: this is not going to work.  Above, you pass an instance of genesys to the initialization function.  Here you are creating a new instance.

we have only one connection to the genesys -- we want only one instance.

Instruments/nmr_signal_gui.py
```diff
@@ -120,26 +126,67 @@ def on_pick(self, event):
...
+            with prologix_connection() as pro_log:
+                with LakeShore475(pro_log) as h:
```
jmfranck: same comment as for genesys

Instruments/nmr_signal_gui.py
```diff
@@ -41,11 +41,17 @@
...
+    LakeShore475,
+    prologix_connection,
+    power_control,
+)
 
 
 class NMRWindow(QMainWindow):
-    def __init__(self, xepr, myconfig, parent=None):
-        self.xepr = xepr
+    def __init__(self, genesys, myconfig, parent=None):
```
jmfranck: actually -- you need a second argument here, where you pass the instance of the hall probe connection.

Instruments/nmr_signal_gui.py
```diff
@@ -120,26 +126,67 @@ def on_pick(self, event):
...
+                            / config_dict["gamma_eff_mhz_g"]
+                        )
+                        ramp_steps = I_setting * 2
+                        ramp_dt = 0.05
+                        for I in np.linspace(0.0, I_setting, ramp_steps):
+                            g.I_limit = I
+                            time.sleep(ramp_dt)
+                        Field = I_setting / config_dict["current_v_field_A_G"]
+                        self.prev_field = Field
+                        print("about to return from set_field_conditional")
```
jmfranck: I'm not really reading the bulk of above -- I assume this is copied + pasted from your ppg, where it works.

Instruments/power_control_server.py
jmfranck, line 1: I'm a strong believer that:

time to review pull request ∝number of lines changed
number of cycles of reviews needed to resolve pull request ∝number of lines changed

time to resolve pull request ∝ (number of lines changed)²

so we need to simplify into as small parts as possible.  Here we can clearly separate changes to power control server from changes to nmr gui.  (also we only need the first for this week!)

So just roll this back:
`git checkout master Instruments/power_control_server.py` will just check out the master version of the file for your next commit.

Instruments/nmr_signal_gui.py
jmfranck: **line 521** is where the instance of `xepr` used to be created. That's where you want to create the instances of both the hall probe and the genesys.

jmfranck: specifically, this will be something like:
```
with genesys(...
    with prologix(...
        with LakeShore(...
```

Note that you can change the order so that genesys comes innermost, but lakeshore MUST come inside the prologix, since you pass it the prologix instance.

Remember -- we make 2 different TCP/IP socket connections to the hall probe vs. the power supply. (They are not related to each other, and don't affect each other)

jmfranck:
note that when everything is said + done, the flake check at the bottom here should pass.
I believe you have my vscode config w/ the ruff extension, etc --> if that's the case then the flake test here should be the same as the errors you get when you press `ctrl-shift-m` in vscode.

Instruments/nmr_signal_gui.py
atahangarip: Done. Do I need another instance for prologix connection in the initilization.

Instruments/nmr_signal_gui.py
```diff
@@ -120,24 +124,57 @@ def on_pick(self, event):
...
+            self.myconfig["current_v_field_A_G"],
+        )
+        self.myconfig["current_v_field_A_G"] *= self.prev_field / true_B0_G
+        print("to", self.myconfig["current_v_field_A_G"])
         if (
             hasattr(self, "prev_field")
-            and abs(Field - self.prev_field) > 50.0 / gammabar_H * 1e4
-            and abs(Field - self.prev_field) < 850.0 / gammabar_H * 1e4
+            and abs(Field - self.prev_field)
+            > 50.0e-6 / self.myconfig["gamma_eff_mhz_g"]
```
jmfranck: the thing on the left is clearly "the field equivalent of 50 Hz" and "the field equivalent of 850 Hz".  I don't understand what this is supposed to be.

Instruments/nmr_signal_gui.py
```diff
@@ -120,24 +124,57 @@ def on_pick(self, event):
         QMessageBox.information(self, "Click!", msg)
 
     def set_field_conditional(self, Field):
+        I_setting = Field * self.myconfig["current_v_field_A_G"]
+        true_B0_G = self.h.field.to("T").magnitude * 1e4
+        print(
+            "adjusting current_v_field_A_G from",
+            self.myconfig["current_v_field_A_G"],
+        )
+        self.myconfig["current_v_field_A_G"] *= self.prev_field / true_B0_G
```
jmfranck: this has the potential to error out -- note that the code below is checking that prev_field actually exists first.

Instruments/nmr_signal_gui.py
```diff
@@ -120,24 +124,57 @@ def on_pick(self, event):
...
+                * (self.myconfig["current_v_field_A_G"])
+                / self.myconfig["gamma_eff_mhz_g"]
+            )
+            self.g.I_limit(I_setting)
+            Field = I_setting / self.myconfig["current_v_field_A_G"]
             self.prev_field = Field
         elif (
             hasattr(self, "prev_field")
-            and abs(Field - self.prev_field) < 50.0 / gammabar_H * 1e4
+            and abs(Field - self.prev_field)
```
jmfranck: again -- I don't understand what the rationale for this change is.

Instruments/nmr_signal_gui.py
```diff
@@ -120,24 +124,57 @@ def on_pick(self, event):
...
+                self.myconfig["carrierfreq_mhz"] / Field
+            )
+            I_setting = (
+                (self.myconfig["carrierFreq_MHz"])
+                * (self.myconfig["current_v_field_A_G"])
+                / self.myconfig["gamma_eff_mhz_g"]
+            )
+            ramp_steps = round(I_setting * 2)
+            ramp_dt = 0.05
+            for I in np.linspace(0.0, I_setting, ramp_steps):
```
jmfranck: this is definitely not the place for a field ramp.  The context here is that this code is adjusting for small differences in order to center the NMR signal at 0 offset.  Which means that any field changes are going to be small.

Instruments/nmr_signal_gui.py
```diff
@@ -226,9 +263,12 @@ def regen_plots(self):
         self.echo_data.ift("t2")
         filter_timeconst = self.apo_time_const
         self.echo_data *= np.exp(
-            -abs((
-                self.echo_data.fromaxis("t2") - self.myconfig["tau_us"] * 1e-6
-            ))
+            -abs(
+                (
+                    self.echo_data.fromaxis("t2")
```
jmfranck: why are you making this change?

Instruments/nmr_signal_gui.py
```diff
@@ -240,9 +280,9 @@ def pyspec_plot(*args, **kwargs):
             myy = args[0]
             longest_dim = np.argmax(myy.data.shape)
             if len(myy.data.shape) > 1:
-                all_but_longest = set(range(len(myy.data.shape))) ^ set((
-                    longest_dim,
-                ))
+                all_but_longest = set(range(len(myy.data.shape))) ^ set(
```
jmfranck: this seems to be a formatting change -- could be from black or something.  But, you are not passing flake anyways, so I woudl just not make this change for now.

Instruments/nmr_signal_gui.py
```diff
@@ -468,8 +508,43 @@ def create_action(
...
+                    g.output = True
+                    print("The power supply is on.")
+                except:
+                    raise TypeError("The power supply is not connected.")
+                # {{{ ramp up the field
+                print("Ramping up the field")
+                for I in np.linspace(0.0, I_setting, ramp_steps):
+                    g.I_limit = I
+                    time.sleep(ramp_dt)
+                print(f"Settling for {settle_initial_s:.0f} s")
```
jmfranck, line 545: printing is good to help diagnose when you're setting up, but it's not good to actually put in the final PR.  We can leave this for now, but need to remove before merging the PR.

PR #44, opened 2025-12-17
Power Control Server code updates

atahangarip:


examples/combined_ODNP.py
```diff
@@ -143,142 +147,14 @@
...
-        filename = "temp_ctrl.h5"
-        control_thermal.hdf5_write(filename, directory=target_directory)
-        final_log.append(
-            "change the name accordingly once this is done running!"
-        )
-# }}}
-logger.info("\n*** FILE SAVED IN TARGET DIRECTORY ***\n")
-logger.debug(psd.strm("Name of saved data", control_thermal.name()))
-# }}}
-# {{{IR at no power
```
jmfranck, line 208: you do want to make sure that it's still acquiring an IR measurement at no power there should be a loop of IR measurements at different powers, and you want to make sure that no power is included there.

examples/active.ini
```diff
@@ -1,26 +1,31 @@
 [acq_params]
 amplitude = 0.2
 p90_us = 1.7
-beta_90_s_sqrtw = 2.5e-05
-deadtime_us = 10.0
-tau_us = 3500.0
+beta_90_s_sqrtw = 8.12e-06
```
jmfranck, line 4: most of these changes are not justified.  We don't want to create noise by just changing lots of the adjustable parameters.

examples/set_field.py
```diff
@@ -0,0 +1,31 @@
...
+    turns the PS off.
+"""
+
+import sys
+from Instruments import power_control
+from SpinCore_pp import configuration
+
+config_dict = configuration("active.ini")
+with power_control() as p:
+    if len(sys.argv) < 2:
```
jmfranck: demand the correct number of arguments.  Here, you mean == 1, not <2.  In the next part, you mean == 2, and at the end, you should have else with a raise error.

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+        Genesys power supply instance with I_limit property.
+    """
+    true_B0_G = h.field_in_G
+    logging.debug(
+        strm(
+            "adjusting current_v_field_A_G from",
+            config_dict["current_v_field_A_G"],
+        )
+    )
+    config_dict["current_v_field_A_G"] *= B0_des_G / true_B0_G
```
jmfranck: this is not quite right. The B0_des_G that should be used in this calcualtion is not just whatever you passed to the function.  RATHER, it's the current current (pun/duplication intended), divided by the current value of current_v_field_A_G

jmfranck: probably do the algebra on that for a simpler expression

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+    #     to where we think we want to
+    #     be, allowing for the
+    #     possibility that it might be a
+    #     large change
+    if I_setting > 25:
+        raise ValueError("Current is too high.")
+    try:
+        if not gen.output:
+            h.zero_probe()
+            logging.info("Calibration for 40s")
```
jmfranck: "calibrating hall probe" --> because in context, you might not know what's being calibrated.

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+    num_field_matches = 0
+    for j in range(30):
+        time.sleep(config_dict["magnet_settle_short"])
+        field_discrepancy = abs(h.field_in_G - B0_des_G)
+        if field_discrepancy > 2.0:
+            time.sleep(config_dict["magnet_settle_medium"])
+        if (
+            # as we approach lower fields, we encounter a no-current
+            # discrepancy that can't be calibrated out.
+            (B0_des_G < 100 and field_discrepancy > 5)
```
jmfranck: it sounded to me like the 100 here should be more like 10.

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+            or (B0_des_G >= 20 and field_discrepancy > 0.8)
```
jmfranck, line 119: 20 should be 10

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+        ):
+            adjust_field(
+                B0_des_G,
+                config_dict,
+                h,
+                gen,
+            )
+            num_field_matches = 0
+        else:
```
jmfranck: there was commentary here about smaller steps that would need to be handled by the shim stack that have been thrown out.  Those are missing, and should be restored.

Instruments/field_feedback.py
```diff
@@ -0,0 +1,140 @@
...
+            " three times in a row, and it"
+            " didn't work!"
+        )
+    # }}}
+    if B0_des_G < 20:
+        gen.I_limit = 0
+        gen.output = False
+        logging.info("The PS is off.")
+    true_B0_G = h.field_in_G
+    logging.info(
```
jmfranck: debug, not info

Instruments/gds.py
```diff
@@ -247,8 +247,10 @@ def upto_hashtag():
         self.read_binary(6)  # length of 550000
         data = self.read_binary(50001)
         assert data[-1] == 10, (
-            "data is not followed by newline!, rather it's %d" % data[-1]
+            "data is not followed by newline!, rather it's %d" % data[-1],
+            f"... length of data {len(data)}"
         )
+        print('length of data',len(data))
```
jmfranck: for this and other reasons, you need to make sure it passes the flake test.

Instruments/power_control_server.py
jmfranck, line 1: this is a complicated diff that we had reviewed together -- should be good to go.

gds_for_tune.py
jmfranck, line 1: this is just garbage from before.  Can you put the version of the file that you are actually using here?

Instruments/field_feedback.py
atahangarip: It is better to make it a bit higher due to field instability. So, I am making it 20.

PR #46, opened 2026-01-26
Combined odnp updates

jmfranck:


jmfranck:
we need to first merge #44 (most of the changes here are covered in #44)

PR #47, opened 2026-01-27
GUI_edits

atahangarip:
Adding a textbox to change gamma_eff_mhz_g in NMR Signal GUI and a spinbox to control MW power in MW Tuning GUI.

### DIFF
--- a/Instruments/bridge12.py
+++ b/Instruments/bridge12.py
@@ -245,15 +245,14 @@ class Bridge12(Serial):
         self.write(b"power %d\r" % setting)
         _ = self.readline()  # gobble the power updated statement
 
-    def set_power(self, dBm, bypass_increment_limit=False):
+    def set_power(self, dBm):
         """set *and check* power.  On successful completion, set
         `self.cur_pwr_int` to 10*(power in dBm).
 
         Need to have 2 safeties for set_power:
 
         1. When power is increased above 10 dBm, the power is not allowed to
-           increase by more than 3 dBm above the current power, unless
-           bypass_increment_limit is True.
+           increase by more than 3 dBm above the current power.
         2. When increasing the power, call the power reading function.
 
         Parameters
@@ -284,9 +283,7 @@ class Bridge12(Serial):
                     "Before you try to set the power above 10 dBm, you must"
                     " first set a lower power!!!"
                 )
-            if (not bypass_increment_limit) and (
-                setting > 30 + self.cur_pwr_int
-            ):
+            if setting > 30 + self.cur_pwr_int:
                 raise RuntimeError(
                     "Once you are above 10 dBm, you must raise the power in"
                     " MAX 3 dB increments.  The power is currently %g, and you"
@@ -612,9 +609,9 @@ class Bridge12(Serial):
                         " lower than the rx_dBm of the dip, which doesn't make"
                         " sense -- check %gdBm_%s" % (10.0, rx_dBm)
                     )
-        assert (
-            self.frq_sweep_10dBm_has_been_run
-        ), "I should have run the 10 dBm curve -- not sure what happened"
+        assert self.frq_sweep_10dBm_has_been_run, (
+            "I should have run the 10 dBm curve -- not sure what happened"
+        )
         over_diff = r_[
             0, np.diff(np.int32(over_bool))
         ]  # should indicate whether this position has lifted over (+1) or
@@ -704,8 +701,8 @@ class Bridge12(Serial):
         p = np.polyfit(freq, rx, 2)
         c, b, a = p
         # polynomial of form a+bx+cx^2
-        self.fit_data[self.last_sweep_name + "_func"] = (
-            lambda x: a + b * x + c * x**2
+        self.fit_data[self.last_sweep_name + "_func"] = lambda x: (
+            a + b * x + c * x**2
         )
         self.fit_data[self.last_sweep_name + "_range"] = freq[r_[0, -1]]
         # the following should be decided from doing algebra (I haven't
--- a/Instruments/bridge12.py
+++ b/Instruments/bridge12.py
@@ -245,14 +245,15 @@ class Bridge12(Serial):
         self.write(b"power %d\r" % setting)
         _ = self.readline()  # gobble the power updated statement
 
-    def set_power(self, dBm):
+    def set_power(self, dBm, bypass_increment_limit=False):
         """set *and check* power.  On successful completion, set
         `self.cur_pwr_int` to 10*(power in dBm).
 
         Need to have 2 safeties for set_power:
 
         1. When power is increased above 10 dBm, the power is not allowed to
-           increase by more than 3 dBm above the current power.
+           increase by more than 3 dBm above the current power, unless
+           bypass_increment_limit is True.
         2. When increasing the power, call the power reading function.
 
         Parameters
@@ -283,7 +284,9 @@ class Bridge12(Serial):
                     "Before you try to set the power above 10 dBm, you must"
                     " first set a lower power!!!"
                 )
-            if setting > 30 + self.cur_pwr_int:
+            if (not bypass_increment_limit) and (
+                setting > 30 + self.cur_pwr_int
+            ):
                 raise RuntimeError(
                     "Once you are above 10 dBm, you must raise the power in"
                     " MAX 3 dB increments.  The power is currently %g, and you"
--- a/Instruments/bridge12.py
+++ b/Instruments/bridge12.py
@@ -499,6 +499,8 @@ class Bridge12(Serial):
         )  # change to the screen that shows the reflection
         rxvalues = np.zeros(len(freq))
         txvalues = np.zeros(len(freq))
+        if not hasattr(self, "cur_pwr_int"):
+            self.cur_pwr_int = self.power_int()
         if not self.frq_sweep_10dBm_has_been_run:
             if self.cur_pwr_int != 100:
                 raise ValueError(
--- a/Instruments/bridge12.py
+++ b/Instruments/bridge12.py
@@ -499,8 +499,6 @@ class Bridge12(Serial):
         )  # change to the screen that shows the reflection
         rxvalues = np.zeros(len(freq))
         txvalues = np.zeros(len(freq))
-        if not hasattr(self, "cur_pwr_int"):
-            self.cur_pwr_int = self.power_int()
         if not self.frq_sweep_10dBm_has_been_run:
             if self.cur_pwr_int != 100:
                 raise ValueError(
--- a/Instruments/microwave_tuning_gui.py
+++ b/Instruments/microwave_tuning_gui.py
@@ -44,7 +44,7 @@ class TuningWindow(qt6w.QMainWindow):
         self.timer.setInterval(100)  # .1 seconds
         self.timer.timeout.connect(self.opt_update_frq)
         self.timer.start(1000)
-        self.last_sweep_power_dbm = self.spinbox_power.value()
+        self.last_sweep_power_dbm = None
         self.on_recapture()
         # self._n_times_run = 0
 
@@ -82,8 +82,7 @@ class TuningWindow(qt6w.QMainWindow):
         qt6w.QMessageBox.about(self, "About the demo", msg.strip())
 
     def on_power_edit(self):
-        power_dbm = self.spinbox_power.value()
-        print("you changed MW power to", power_dbm, "dBm")
+        print("you changed MW power to", self.spinbox_power.value(), "dBm")
         return
 
     def orig_zoom_limits(self):
@@ -172,14 +171,21 @@ class TuningWindow(qt6w.QMainWindow):
         return
 
     def on_recapture(self):
-        requested_power_dbm = self.spinbox_power.value()
-        prev_sweep_power_dbm = self.last_sweep_power_dbm
-        power_diff = requested_power_dbm - prev_sweep_power_dbm
+        power_diff = self.spinbox_power.value() - (
+            self.last_sweep_power_dbm
+            if self.last_sweep_power_dbm is not None
+            else self.spinbox_power.value()
+        )
+        """
+        In the future, we can choose threshold of a "reasonable reflection"
+        value based on the current curve and limits and adjust the value of
+        the reasonable power increment steps.
+        """
         if power_diff > 3.0:
             reply = qt6w.QMessageBox.question(
                 self,
                 "Confirm power change",
-                f"I am setting the power to {requested_power_dbm}"
+                f"I am setting the power to {self.spinbox_power.value()}"
                 "dBm. Are you sure?",
                 qt6w.QMessageBox.Yes | qt6w.QMessageBox.No,
                 qt6w.QMessageBox.No,
@@ -188,11 +194,15 @@ class TuningWindow(qt6w.QMainWindow):
                 # Revert to previous value
                 self.textbox_power.setValue(self._last_power_dbm)
                 return
-        self.B12.set_power(requested_power_dbm, bypass_increment_limit=True)
+        for power_val in np.linspace(
+            self.spinbox_power.value() - power_diff,
+            self.spinbox_power.value(),
+            int(power_diff / 3) + 1,
+        ):
+            self.B12.set_power(power_val)
         self.generate_data()
         self.regen_plots()
-        self.B12.set_power(prev_sweep_power_dbm, bypass_increment_limit=True)
-        self.last_sweep_power_dbm = requested_power_dbm
+        self.last_sweep_power_dbm = self.spinbox_power.value()
         return
 
     def regen_plots(self):
--- a/Instruments/microwave_tuning_gui.py
+++ b/Instruments/microwave_tuning_gui.py
@@ -32,6 +32,8 @@ class TuningWindow(qt6w.QMainWindow):
         self.setWindowTitle("B12 tuning!")
         self.setGeometry(20, 20, 1500, 800)
 
+        self.last_sweep_power_dBm = None
+        self.target_power_dBm = 10.0
         self.create_menu()
         self.create_main_frame()
         self.create_status_bar()
@@ -44,9 +46,6 @@ class TuningWindow(qt6w.QMainWindow):
         self.timer.setInterval(100)  # .1 seconds
         self.timer.timeout.connect(self.opt_update_frq)
         self.timer.start(1000)
-        self.last_sweep_power_dBm = None
-        self.target_power_dBm = 10.0
-        self.on_power_edit(10.0)
         self.on_recapture()
         # self._n_times_run = 0
 
@@ -339,7 +338,9 @@ class TuningWindow(qt6w.QMainWindow):
         self.spinbox_power.setRange(0, 40.0)
         self.spinbox_power.setSingleStep(1.0)
         self.spinbox_power.setValue(10.0)
-        self.spinbox_power.editingFinished.connect(self.on_power_edit)
+        self.spinbox_power.editingFinished.connect(
+            self.on_power_edit(self.target_power_dBm)
+        )
         self.textboxes_vbox.addWidget(self.power_label)
         self.textboxes_vbox.addWidget(self.spinbox_power)
         # }}}
--- a/Instruments/microwave_tuning_gui.py
+++ b/Instruments/microwave_tuning_gui.py
@@ -87,7 +87,8 @@ class TuningWindow(qt6w.QMainWindow):
 
     @property
     def target_power_dbm(self):
-        return self.target_power_dbm
+        target_power = self.spinbox_power.value()
+        return target_power
 
     @target_power_dbm.setter
     def target_power_dbm(self, power_dbm):
@@ -200,12 +201,12 @@ class TuningWindow(qt6w.QMainWindow):
                 )
                 if reply != qt6w.QMessageBox.Yes:
                     # Revert to previous value
-                    self.textbox_power.setValue(self._last_power_dbm)
+                    self.target_power_dbm = self.last_sweep_power_dbm
                     return
             for power_val in np.linspace(
                 self.last_sweep_power_dbm,
                 self.target_power_dbm,
-                int(power_diff / 3) + 1,
+                int(abs(power_diff) / 3) + 1,
             ):
                 self.B12.set_power(power_val)
         self.generate_data()
--- a/Instruments/microwave_tuning_gui.py
+++ b/Instruments/microwave_tuning_gui.py
@@ -179,7 +179,7 @@ class TuningWindow(qt6w.QMainWindow):
             reply = qt6w.QMessageBox.question(
                 self,
                 "Confirm power change",
-                f"I am setting the power to {power_dbm} dBm. Are you sure?",
+                f"I am setting the power to {requested_power_dbm} dBm. Are you sure?",
                 qt6w.QMessageBox.Yes | qt6w.QMessageBox.No,
                 qt6w.QMessageBox.No,
             )
--- a/Instruments/microwave_tuning_gui.py
+++ b/Instruments/microwave_tuning_gui.py
@@ -82,7 +82,16 @@ class TuningWindow(qt6w.QMainWindow):
         qt6w.QMessageBox.about(self, "About the demo", msg.strip())
 
     def on_power_edit(self):
-        print("you changed MW power to", self.spinbox_power.value(), "dBm")
+        print("you changed MW power to", self.target_power_dbm, "dBm")
+        return
+
+    @property
+    def target_power_dbm(self):
+        return self.target_power_dbm
+
+    @target_power_dbm.setter
+    def target_power_dbm(self, power_dbm):
+        self.spinbox_power.setValue(power_dbm)
         return
 
     def orig_zoom_limits(self):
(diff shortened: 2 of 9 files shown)


Instruments/microwave_tuning_gui.py
```diff
@@ -80,6 +80,38 @@ def on_about(self):
         """
         qt6w.QMessageBox.about(self, "About the demo", msg.strip())
 
+    def on_power_edit(self):
+        power_dbm = self.textbox_power.value()
```
jmfranck: Change the name of this to spinbox_power since it should be a spinbox.

Instruments/microwave_tuning_gui.py
```diff
@@ -80,6 +80,38 @@ def on_about(self):
...
+        if power_dbm > 35.0:
```
jmfranck: Don't do this.  I think it actually goes up to 40 dBm.   We tend not to use more because we don't want to overheat the sample, but we don't want to hard code in a limit

Instruments/microwave_tuning_gui.py
```diff
@@ -80,6 +80,38 @@ def on_about(self):
...
+                qt6w.QMessageBox.No,
+            )
+            if reply != qt6w.QMessageBox.Yes:
+                # Revert to previous value
+                self.textbox_power.setValue(self._last_power_dbm)
+                return
+
+        print("you changed MW power to", power_dbm, "dBm")
+        self.B12.set_power(power_dbm)
+        self._last_power_dbm = power_dbm
```
jmfranck: Do not do this here... Move this to after we have conducted a sweep.

We don't really know if the cavity is reasonably tuned at the last power until we have actually conducted a sweep.

Instruments/microwave_tuning_gui.py
```diff
@@ -297,6 +329,18 @@ def create_main_frame(self):
         self.orig_zoom_limits()
         # }}}
 
+        # {{{ box to set MW power
+        self.power_label = qt6w.QLabel("MW Power (dBm)")
+        self.textbox_power = qt6w.QDoubleSpinBox()
+        self.textbox_power.setDecimals(2)
```
jmfranck: The B12 only has a power resolution of 0.1 dBm (zero point it might actually be 0.5, but I think it's 0.1). In any case, the point is that we are not able to use two decimal places

Instruments/microwave_tuning_gui.py
```diff
@@ -297,6 +329,18 @@ def create_main_frame(self):
...
+        self.textbox_power.setRange(-100.0, 35.0)
+        self.textbox_power.setSingleStep(0.5)
```
jmfranck: If I'm correct in assuming this is how much it changes when you click on the arrows, I would just set this to 1.0

Instruments/microwave_tuning_gui.py
```diff
@@ -297,6 +329,18 @@ def create_main_frame(self):
         self.orig_zoom_limits()
         # }}}
 
+        # {{{ box to set MW power
+        self.power_label = qt6w.QLabel("MW Power (dBm)")
+        self.textbox_power = qt6w.QDoubleSpinBox()
+        self.textbox_power.setDecimals(2)
+        self.textbox_power.setRange(-100.0, 35.0)
```
jmfranck: This should be 0 to 40.  It's not that zero is special, but I don't think it does lower than zero.

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
         self.myconfig["beta_90_s_sqrtW"] = float(thetext)
         return
 
+    def on_gamma_edit(self):
+        thetext = self.textbox_gamma.text()
+        print("you changed gamma_eff_MHz_G to", thetext)
+        self.myconfig["gamma_eff_MHz_G"] = float(thetext)
```
jmfranck: Here you would also have a function that would update the position of the axvline

jmfranck: Although maybe you want to leave this for a subsequent PR. Up to you

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
         # matplotlib.backend_bases.PickEvent
         #
         # It carries lots of information, of which we're using
         # only a small amount here.
         #
+        if (
+            isinstance(event.artist, Line2D)
+            and event.artist is self.centerline
+        ):
+            return
```
jmfranck, line 157: This code here does not seem good. I would roll this back

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
         box_points = event.artist.get_bbox().get_points()
```
jmfranck: This is because I had adapted this originally from some example where there were multiple bars, and this is pulling the bounding box for the bars. Clearly, that is not what we want to do.

However, I do believe it is possible to have it return the nearest coordinates of the line that you are graphing, which I do think you want it to do.

If you are considering a subsequent PR, then this would be grouped with the business of updating the axvline

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
         msg = "You've clicked on a bar with coords:\n %s" % box_points
 
         QMessageBox.information(self, "Click!", msg)
 
+    def on_center_press(self, event):
+        if not self.dragcenter_cb.isChecked():
```
jmfranck: Remove this conditional... You shouldn't have to check a box to be able to drag the line.

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
+            isinstance(event.artist, Line2D)
+            and event.artist is self.centerline
+        ):
+            return
         box_points = event.artist.get_bbox().get_points()
         msg = "You've clicked on a bar with coords:\n %s" % box_points
 
         QMessageBox.information(self, "Click!", msg)
 
+    def on_center_press(self, event):
```
jmfranck, line 145: I'm going through top to bottom, so keep that in mind as you are reading my comments... Comments above were made before I saw this

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
+            return
+        if event.inaxes != self.axes:
+            return
+        if self.centerline is None:
+            return
+        if self.centerline.axes is None:
+            return
+        if self.centerline.figure is None:
+            return
+        if event.button != 1:
```
jmfranck: I do not know what this means. Add a comment to explain

jmfranck: I do not know what this means. Add a comment to explain

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
+            return
+        if event.xdata is None:
```
jmfranck: I would group all these conditionals onto a single line (inside parens, so it will break across multiple lines but the point is to not have five return statements) with or.

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
+            return
+        self._dragging_center = True
+
+    def on_center_motion(self, event):
+        if not self._dragging_center:
+            return
+        if event.inaxes != self.axes or event.xdata is None:
+            return
+        self.centerfrq_override = event.xdata
+        self.centerline.set_xdata([event.xdata, event.xdata])
```
jmfranck: Okay... You do not want two separate PRs.

This line should go into a helper function (in general, here when I'm proposing new functions, I'm implying that they are methods of the class) that accepts what is x data here as an argument.

This helper function will execute this line as well as the subsequent canvas redraw. 

We will use it here and elsewhere

Instruments/nmr_signal_gui.py
```diff
@@ -122,18 +129,67 @@ def on_plen_edit(self):
...
+        self.canvas.draw_idle()
+
+    def on_center_release(self, event):
+        if not self._dragging_center:
+            return
+        self._dragging_center = False
+        if self.centerfrq_override is not None:
```
jmfranck: We do not want this conditional. Whenever we drag and release the center line, we do want to update the values. This should not be checkbox dependent. 

Here you want to call: 

- The helper function defined above, which updates the position of the axvline. (To be sure that we end at the actual position where we released the drag)

- A second helper function responsible for calculating and setting the new gamma effective value based on the offset value (something like this may actually exist already and you want to double check ... It's possible the code exists, but it is not yet grouped as a function that we can reuse it)

- executing a call back that changes the value in the gamma effective spin box. I believe that this will also invoke the callbacks associated with the spin box, so that the actual value of gamma effective in the config dictionary gets changed as well.

You also want to do the reverse, which is that when you change the value of gamma effective, you want to calculate the offset that you expect from the new gamma effective, and then call the first of the two helper functions here to update the position of the axvline.  In order to avoid an infinite loop, I think that this means you will want a private (means starting with an _) attribute that indicates whether or not you are manipulating the axvline by clicking on the plot.

Instruments/nmr_signal_gui.py
jmfranck: Yes, so just delete this entire function and any callbacks that use it.

jmfranck: Because I think your click and drag mechanism below is better.

Instruments/nmr_signal_gui.py
```diff
@@ -327,7 +383,15 @@ def pyspec_plot(*args, **kwargs):
                         alpha=0.2,
                     )
         centerfrq = signal.C.argmax("t2").item()
-        self.axes.axvline(x=centerfrq, ls=":", color="r", alpha=0.25)
+        if (
+            self.dragcenter_cb.isChecked()
+            and self.centerfrq_override is not None
```
jmfranck: I am stopping here with this review, but will note that you should also read these parts keeping in mind the comments from above... For example, there is a pattern where we do not want to have to click a box in order to manipulate the gamma effective. Just clicking and dragging the axvline should be sufficient.

Also, I do not know what exactly you mean by frequency override... Keep in mind that the purpose of the axvline is that it shows us the position in the spectrum that will show up at zero offset *after* we acquire a spectrum with the value of gamma effective that we have either allowed the program to choose automatically for us, or have manually entered into the spin box.

jmfranck (changes requested):

Instruments/microwave_tuning_gui.py
```diff
@@ -80,6 +80,38 @@ def on_about(self):
...
+                self,
+                "Confirm power change",
+                f"I am setting the power to {power_dbm} dBm. Are you sure?",
+                qt6w.QMessageBox.Yes | qt6w.QMessageBox.No,
+                qt6w.QMessageBox.No,
+            )
+            if reply != qt6w.QMessageBox.Yes:
+                # Revert to previous value
+                self.textbox_power.setValue(self._last_power_dbm)
+                return
```
atahangarip: 3 dBm increments are already not allowed by Bridge12 so I am removing this block

Instruments/nmr_signal_gui.py
atahangarip: It checks which mouse button is used and only proceeds only when the left button is used. So, the main function is to ignore any left clicks.

Instruments/microwave_tuning_gui.py
atahangarip: I added an argument in Bridge12.py to bypass that and put the msg block again.

Instruments/bridge12.py
```diff
@@ -245,14 +245,15 @@ def calit_power(self, dBm):
         self.write(b"power %d\r" % setting)
         _ = self.readline()  # gobble the power updated statement
 
-    def set_power(self, dBm):
+    def set_power(self, dBm, bypass_increment_limit=False):
```
jmfranck: do not do this.

jmfranck: if you want to increment by more than 3, you HAVE to do it stepwise.  This is the point.

jmfranck: it's doing this to give the code and the B12 a chance to trigger interlocks regarding the reflected power, before the reflected power becomes high enough to damage things.

jmfranck: my philosophy with this is that the fact that you did this means the **docstring is not clear enough**, so edit the docstring slightly to explain this.

Instruments/bridge12.py
```diff
@@ -245,14 +245,15 @@ def calit_power(self, dBm):
...
         """set *and check* power.  On successful completion, set
         `self.cur_pwr_int` to 10*(power in dBm).
 
         Need to have 2 safeties for set_power:
 
         1. When power is increased above 10 dBm, the power is not allowed to
-           increase by more than 3 dBm above the current power.
+           increase by more than 3 dBm above the current power, unless
+           bypass_increment_limit is True.
```
jmfranck: following comments above, edit docstring here.

Instruments/bridge12.py
```diff
@@ -283,7 +284,9 @@ def set_power(self, dBm):
                     "Before you try to set the power above 10 dBm, you must"
                     " first set a lower power!!!"
                 )
-            if setting > 30 + self.cur_pwr_int:
+            if (not bypass_increment_limit) and (
+                setting > 30 + self.cur_pwr_int
+            ):
```
jmfranck: roll back

Instruments/microwave_tuning_gui.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, B12, config, parent=None):
         self.timer.setInterval(100)  # .1 seconds
         self.timer.timeout.connect(self.opt_update_frq)
         self.timer.start(1000)
+        self.last_sweep_power_dbm = self.spinbox_power.value()
```
jmfranck: when starting, last_sweep_power_dBm should be `None` b/c no sweep has been performed.

Instruments/microwave_tuning_gui.py
```diff
@@ -80,6 +81,11 @@ def on_about(self):
         """
         qt6w.QMessageBox.about(self, "About the demo", msg.strip())
 
+    def on_power_edit(self):
+        power_dbm = self.spinbox_power.value()
+        print("you changed MW power to", power_dbm, "dBm")
```
jmfranck: combine these to one line.  (just a stylistic thing --> in general, I do not like spinning up variables that are used on the next line, unless we're doing some really complicated algebra and it makes it more legible.

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +172,27 @@ def generate_data(self):
         return
 
     def on_recapture(self):
+        requested_power_dbm = self.spinbox_power.value()
+        prev_sweep_power_dbm = self.last_sweep_power_dbm
+        power_diff = requested_power_dbm - prev_sweep_power_dbm
+        if power_diff > 3.0:
+            reply = qt6w.QMessageBox.question(
```
jmfranck: we should make a future note here.  Based on the current curve, the limits that we have chosen, and a threshold of a "reasonable reflection", we can actually predict how much of an increase is reasonable.

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +172,27 @@ def generate_data(self):
...
+                f"I am setting the power to {requested_power_dbm}"
+                "dBm. Are you sure?",
+                qt6w.QMessageBox.Yes | qt6w.QMessageBox.No,
+                qt6w.QMessageBox.No,
+            )
+            if reply != qt6w.QMessageBox.Yes:
+                # Revert to previous value
+                self.textbox_power.setValue(self._last_power_dbm)
+                return
+        self.B12.set_power(requested_power_dbm, bypass_increment_limit=True)
```
jmfranck: w.r.t. comments above -- here, e.g. you would loop in steps of 3 dB

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +172,27 @@ def generate_data(self):
...
         self.generate_data()
         self.regen_plots()
+        self.B12.set_power(prev_sweep_power_dbm, bypass_increment_limit=True)
```
jmfranck: I don't understand this.

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +172,27 @@ def generate_data(self):
         return
 
     def on_recapture(self):
+        requested_power_dbm = self.spinbox_power.value()
+        prev_sweep_power_dbm = self.last_sweep_power_dbm
```
jmfranck: again, don't spin up a variable to consume only once

Instruments/microwave_tuning_gui.py
jmfranck: why are you setting back to the previous power?

Instruments/nmr_signal_gui.py
```diff
@@ -103,8 +109,10 @@ def on_about(self):
     def set_default_choices(self):
         self.textbox_apo.setText("10 ms")
         self.textbox_plen.setText("%g" % self.myconfig["beta_90_s_sqrtW"])
+        self.textbox_gamma.setText("%g" % self.myconfig["gamma_eff_MHz_G"])
         self.textbox_apo.setMinimumWidth(10)
         self.textbox_plen.setMinimumWidth(10)
+        self.textbox_gamma.setMinimumWidth(10)
```
jmfranck, line 114: it's a stylistic choice, but I find it makes more sense to group the things by widget.

Instruments/nmr_signal_gui.py
```diff
@@ -122,17 +130,96 @@ def on_plen_edit(self):
...
-        # The event received here is of the type
-        # matplotlib.backend_bases.PickEvent
-        #
-        # It carries lots of information, of which we're using
-        # only a small amount here.
-        #
-        box_points = event.artist.get_bbox().get_points()
-        msg = "You've clicked on a bar with coords:\n %s" % box_points
+    def on_gamma_edit(self):
+        if self._updating_gamma:
```
jmfranck, line 137: add a docstring.

Instruments/nmr_signal_gui.py
```diff
@@ -122,17 +130,96 @@ def on_plen_edit(self):
...
+            or self.centerline.axes is None  # Line is attached to axes
+            or self.centerline.figure is None  # Line is attached to figure
+            or event.xdata is None  # xdata exists
+            or event.button != 1  # Ignore non-left mouse clicks
+        ):
+            return
+
+        line_x = self.centerline.get_xdata()[0]
+        x0, x1 = self.axes.get_xlim()
+        tol = 0.03 * abs(x1 - x0)
```
jmfranck:
```
tol = 0.03*abs(np.diff(self.axes.get_xlim()).item())
```

Instruments/nmr_signal_gui.py
```diff
@@ -122,17 +130,96 @@ def on_plen_edit(self):
...
+        if abs(event.xdata - line_x) > tol:
```
jmfranck: expression for `line_x` here.

Instruments/nmr_signal_gui.py
```diff
@@ -122,17 +130,96 @@ def on_plen_edit(self):
...
+            return
+        self._dragging_center = True
+
+    # Setting target centerline position.
+    def _update_centerline(self, centerfrq):
```
jmfranck: Make
```
@property
def centerfrq(self):
    return self.centerline.get_xdata()[0]
````
Then find and replace above expression (one instance) with self.centerfrq

Take the code for this function (`_update_centerline`) and put it inside this:
```
@centerfrq.setter
def centerfrq(self, centerfrq):
```
then
```
:s/\._update_centerline(\(.*\))/.centerfrq_target = \1/g
```
then `/\<centerfrq`:
- You should be able to remove all references to `centerfrq_target`, including here.

Comment (jmfranck) [reply Instruments/nmr_signal_gui.py]: **this was getting complicated, so I'm just doing it.  review and see that it makes sense**

File: Instruments/microwave_tuning_gui.py (hunk 25 above)
Comment (atahangarip) [reply Instruments/microwave_tuning_gui.py]: Because in the previous pull requrest, one of the comments was to not to set power without looking into the reflection curve. So, I thought that it is reasonable to set to the previous power at the end of the sweep so that it returns to a power that we know the reflection is reasonable. Otherwise, it might be stay at power with high reflection.

Comment (atahangarip) [reply Instruments/microwave_tuning_gui.py]: However, it is easier to loop without doing that if you don't find it necessary. So, I am rolling it back.

File: Instruments/microwave_tuning_gui.py [hunk 32]
```diff
@@ -80,6 +81,20 @@ def on_about(self):
         """
         qt6w.QMessageBox.about(self, "About the demo", msg.strip())
 
+    def on_power_edit(self):
```
jmfranck: right now, this doesn't do anything but print.  I think taht you want to actually call a function of the bridge12 class in order to set the power, right?  Also, this is where you could break a greater than 3 dB change into a stepwise change by 3 dB

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +181,37 @@ def generate_data(self):
         return
 
     def on_recapture(self):
+        if self.last_sweep_power_dbm is None:
```
jmfranck: in keeping with my previous comment -- I think just move this up.  We want to change the power when we actually ask for the power change, not when we ask for a recapture.

jmfranck: so, moving this up would directly satify my previous comments.  I don't do this to give you a chance to think about it, and find somemthing that I'm misssing, if I'm missing anything.

Instruments/microwave_tuning_gui.py
```diff
@@ -166,8 +181,37 @@ def generate_data(self):
...
+                    return
+            for power_val in np.linspace(
+                self.last_sweep_power_dbm,
+                self.target_power_dbm,
+                int(abs(power_diff) / 3) + 1,
+            ):
+                self.B12.set_power(power_val)
         self.generate_data()
         self.regen_plots()
+        self.last_sweep_power_dbm = self.target_power_dbm
```
jmfranck: I think this would move inside `generate_data` since, that is where you actually perform the sweep?

Instruments/microwave_tuning_gui.py
```diff
@@ -182,6 +226,7 @@ def regen_plots(self):
         if self.fmode:
             if not self._already_fmode:
                 self.B12.set_freq(self.dip_frq_GHz * 1e9)
+                self.B12.set_power(self.target_power_dbm)
```
jmfranck: don't change the power here -- see other comments.

Instruments/microwave_tuning_gui.py
atahangarip: The sweep is performed in generate_data (line 176). This function takes the sweep results (self.line_data) and draws. I think since the axes are cleared in the beggining, not moving this one might be better.

jmfranck: I think I'm not understanding your point.  To me, the attribute tracks what the last sweep run was -- so it makes sense to me to set it right after we run the sweep.  But I don't understand the reference to clearing the axes, so maybe you are thinking about this in a different way -- explain? Or save for a conv.

Instruments/microwave_tuning_gui.py
jmfranck: **NOTE** In contrast to what you are  saying on slack, I would interact with the B12 inside *this* function, and leave your property alone.

That way, when we want a change (whether in the code or manually), we change the spinbox.

but, we are relying on the spinbox callback to actually execute that change.

jmfranck: Note that this is similar to what I was advocating with the `axvline` --> have the GUI representation correspond to actual physical reality, and then make a property that makes it easy to mess with the GUI representation.

jmfranck: (there is a slight difference, b/c the gammaa effective is something in our config file, not exactly a "physical reality", where here, we are actually talking about the setting inside the microcontroller of the B12)

Instruments/microwave_tuning_gui.py
```diff
@@ -297,6 +332,18 @@ def create_main_frame(self):
...
         # }}}
 
+        # {{{ box to set MW power
+        self.power_label = qt6w.QLabel("MW Power (dBm)")
+        self.spinbox_power = qt6w.QDoubleSpinBox()
+        self.spinbox_power.setDecimals(1)
+        self.spinbox_power.setRange(0, 40.0)
+        self.spinbox_power.setSingleStep(1.0)
+        self.spinbox_power.setValue(10.0)
+        self.spinbox_power.editingFinished.connect(self.on_power_edit)
```
atahangarip: The second line here does not automatically calls self.on_power_edit since the editing is done with setValue() function. Then, without setting the power in __init__ part, the script doesn't set the initial power and I get an error due to lines 502-507 in bridge.12 py. Therefore, I, now, set the power in the initialization. I didn't use on_power_edit function because we added the RuntimeError in the beggining.

Instruments/microwave_tuning_gui.py
```diff
@@ -44,6 +45,8 @@ def __init__(self, B12, config, parent=None):
         self.timer.setInterval(100)  # .1 seconds
         self.timer.timeout.connect(self.opt_update_frq)
         self.timer.start(1000)
+        self.target_power_dBm = 10.0
+        self.B12.set_power(10.0)
```
atahangarip, line 49: I had to set power in init process otherwise I get an error due to lines 503-507 in bridge12.py

PR #48, opened 2026-02-06
Hp power supply

jmfranck:


examples/test_shim_connections.py
```diff
@@ -4,34 +4,80 @@
...
-        print(HP.check_output(1))
-        print(HP.check_output(2))
-        print(HP.check_output(3))
-        print("*** *** ***")
+    with HP6623A(prologix_instance=p, address=3) as HP1:
+        with HP6623A(prologix_instance=p, address=5) as HP2:
+            HP1.safe_current_on_enable = 1.8
+            HP2.safe_current_on_enable = 1.8
+            print("*** *** ***")
+            HP_list = [
```
atahangarip: Make a dictionary to give the names of Z_0.. shims. Create the instance for the PS (1,2) and the channel (0,1,2). Define a convinience function above.

Instruments/HP6623A.py
```diff
@@ -1,23 +1,262 @@
-from pylab import *
 from .gpib_eth import gpib_eth
 from .log_inst import logger
+import time
 
-class HP6623A (gpib_eth):
+
+class channel_proxy:
```
jmfranck: move channel proxy and property defs to a different module (move together)

Instruments/HP6623A.py
```diff
@@ -1,23 +1,262 @@
...
+
+    def __repr__(self):
+        name = self._prop._name or "<unnamed>"
+        return (
+            f"<channel_proxy {name} bound to {type(self._owner).__name__}"
+            f" at {hex(id(self._owner))}>"
+        )
+
+
+class channel_property:
```
jmfranck: this goes with channel proxy

Instruments/HP6623A.py
```diff
@@ -1,23 +1,262 @@
...
+        self._fset = fset
+        return self
+
+
+class HP6623A(gpib_eth):
     def __init__(self, prologix_instance=None, address=None):
-        r"""initialize a new `HP6623A` power supply class
-        """
-        super().__init__(prologix_instance,address)
+        r"""initialize a new `HP6623A` power supply class"""
```
jmfranck: docstring with parameters

Instruments/HP6623A.py
```diff
@@ -74,51 +355,321 @@ def get_current(self, ch):
...
+        self.write("OCP? %s" % str(ch + 1))
+        return int(float(self.read()))
+
+    def reset_overcurrent(self, ch):
+        """Reset output after overcurrent protection (OCRST)."""
+        self.write("OCRST %s" % str(ch + 1))
+        return
+
+    def set_unmask(self, ch, mask):
+        """Set mask register for channel (UNMASK)."""
```
jmfranck: what is this? can you add more description?

Instruments/HP6623A.py
```diff
@@ -74,51 +355,321 @@ def get_current(self, ch):
...
+        self.write("SRQ %s" % str(setting))
+        return
+
+    def get_srq(self):
+        """Query SRQ setting (SRQ?)."""
+        self.write("SRQ?")
+        return int(float(self.read()))
+
+    def set_pon(self, enable):
+        """Enable/disable power-on SRQ (PON)."""
```
jmfranck: what is this?  can you add more description?

Instruments/HP6623A.py
```diff
@@ -74,51 +355,321 @@ def get_current(self, ch):
...
+        return int(float(self.read()))
+
+    def error(self):
+        """Query error register (ERR?)."""
+        self.write("ERR?")
+        return int(float(self.read()))
+
+    def idn(self):
+        """Query identification string (ID?)."""
+        self.write("ID?")
```
jmfranck, line 403: is there really no star?  Many of the common/shared commands, like idn, tend to have a star.  Just double-check the manual.

Instruments/HP6623A.py
```diff
@@ -74,51 +355,321 @@ def get_current(self, ch):
...
+        self.write("DCPON %s" % str(mode))
+        return
+
+    def rom(self):
+        """Query firmware revision (ROM?)."""
+        self.write("ROM?")
+        return self.read()
+
+    def vmux(self, ch, input_num):
+        """Query analog multiplexer input (VMUX?)."""
```
jmfranck: more info?

Instruments/HP6623A.py
```diff
@@ -74,51 +355,321 @@ def get_current(self, ch):
...
+            # set voltage and current to 0 and turn off set_output on all
+            # channels, before exiting
+            self.set_voltage(i, 0)
+            self.set_current(i, 0)
+            self.set_output(i, 0)
         super().close()
+        return
+
+    @channel_property
+    def voltage(self, channel):
```
jmfranck: consider renaming these to V_limit and I_limit (or something similar).  It's up to you.  I find the "limit" less confusing b/c it reminds that the actual value depends on whether you are V or I limited.

examples/test_shim_connections.py
```diff
@@ -4,34 +4,80 @@
...
+    # Z1_shim = HP_list[1]
+    # Z2_shim = HP_list[2]
+    # X_shim = HP_list[3]
+    # Y_shim = HP_list[4]
+    if output:
+        for index in range(len(HP_list)):
+            this_shim = HP_list[index]
+            if this_shim[-1] == 0.0:
+                print("zero")
+                this_shim[0].current[this_shim[1]] = 0.0
```
jmfranck: This can use some dramatic editing/simplification using the helper function that we discussed yesterday (on slack)

tests/test_hp6623a_channel_property.py
```diff
@@ -0,0 +1,167 @@
+import importlib.util
+import os
+import pathlib
+import sys
+import types
+import unittest
+
+import numpy as np
+
+# Create a minimal Instruments package so we can load only the needed modules.
```
jmfranck, line 10: the point of this is that it should be possible to pytest tests/test_hp6623a_channel_property.py and it should run without skipping if the instrument is attached.

Instruments/HP6623A.py
atahangarip: There is no stars -- page 66 in the manual

examples/test_shim_connections.py
```diff
@@ -2,36 +2,57 @@
...
-        print(HP.output(1, False))
-        print(HP.output(2, 0))
-        print(HP.output(3, 0))
-        print(HP.check_output(1))
-        print(HP.check_output(2))
-        print(HP.check_output(3))
-        print("*** *** ***")
+    with HP6623A(prologix_instance=p, address=3) as HP1:
+        with HP6623A(prologix_instance=p, address=5) as HP2:
+            HP1.safe_current_on_enable = 1.8
```
jmfranck: I don't understand exactly what this does, and a quick glance at the code doesn't really help.

Instruments/HP6623A.py
```diff
@@ -74,51 +166,467 @@ def get_current(self, ch):
...
+                f"max safe current value 1.8 A"
+            )
+        if value == 0:
+            self.set_current(channel, 0)
+            if self._known_output_state[channel] == 1:
+                self.set_output(channel, 0)
+            return
+
+        if self._known_output_state[channel] == 0:
+            if abs(value) > self.safe_current_on_enable:
```
jmfranck: Here it seems to be some type of safety setting, but why is it implemented at the property level, rather than the level of the function called by the property (so that it's global).  Also, in the initialization, you set it to zero.  It seems like the better strategy mwould be to set it to None on initialization, and you could emit a special error message when it is never set.

atahangarip: There were these two todo in the first PR: "    # TODO ☐: make analogous code for current.  Enforce a software interlock
    #         that if we are turning on the output for the first time, the
    #         current limit must be small (maybe just 0 to be safe).  This
    #         prevents us from slamming a somewhat reasonable voltage across a
    #         low-impedance load.
    # TODO ☐: After previous, write a basic example to use the current limit,
    #         with voltage limit set high, and test on the instrument
    #         (generally we will want to control the current, which is propto
    #         field)"
So to be able to change the current interlock limit later I wrote it in the property level and I did not initilize with None due to this "current limit must be small (maybe just 0 to be safe)" line. Should I change it as you described in the comment or keep as it is?

Instruments/HP6623A.py
```diff
@@ -57,13 +133,50 @@ def set_current(self, ch, val):
...
+                0 to 50.5 V -- 0.05 to 0.824 A
         Returns
         =======
         None
-        
+
         """
-        self.write("ISET %s,%s"%(str(ch),str(val)))
+        self.check_if_allowed("I", ch, val)
+        if self.safe_current_on_enable is None:
```
jmfranck: As noted before, I moved the safe current here b/c it's at a low level.  All the other fancy properties, etc, use this, so this is the best place to put it.

Instruments/HP6623A.py
```diff
@@ -26,15 +82,34 @@ def set_voltage(self, ch, val):
...
+                0.015 V of resolution
+                0 to 20.2 V -- 0.05 to 2.06 A
+                0 to 50.5 V -- 0.05 to 0.824 A
         Returns
         =======
         None
-        
         """
-        self.write("VSET %s,%s"%(str(ch),str(val)))
+        self.check_if_allowed("V", ch, val)
```
jmfranck: the previous behavior silently rounded the input voltages.  We don't want to do that, because we're behaving differently vs. what the operator expects.  I changed it so that we straight up refuse to set a value that's not an allowed value, so that the user is aware of the rounding.  We supply the rounding function so that they can round and see for themselves.  This will become important, e.g. when trying to line-search a shim value.

tests/test_hp6623a_channel_property.py
```diff
@@ -0,0 +1,194 @@
...
+        assert all(list(self.hp.V_limit)[0:3] == setval)
+
+    def test_direct_numpy_vector_set(self):
+        """Exercise direct vector assignment across all channels."""
+        self.require_channels(3)
+        setval = np.array([0.3, 0.302, 0.299])
+        self.hp.V_limit = setval
+        self.assertEqual(self.hp.V_limit, setval)
+
+    def test_non_allowed_value_raises(self):
```
jmfranck: I did generate this with codex, but I specifically asked it to add to the same testing class, to keep things simple. I also described the exact tests that I wanted, which were geared around this strategy described above of refusing to set non-allowed values.

Instruments/HP6623A.py
```diff
@@ -57,13 +134,54 @@ def set_current(self, ch, val):
         ch : int
             Channel 1, 2, or 3
         val : float
-            Current you want to set in Amps; check manual for limits on each channel.
+            Current you want to set in Amps.
+            Channel 1:
```
jmfranck, line 228: rather than saying this, note that these are described by the attributes.

Instruments/HP6623A.py
jmfranck: (because I edited after this comment, this now refers to a few lines down, where the safe current is mentioned)

jmfranck:
fully merged by PR #50

PR #50, opened 2026-03-05
Z0 shim linear field adjustment

atahangarip:


Instruments/field_feedback.py
```diff
@@ -4,6 +4,52 @@
...
+        Desired magnetic field in Gauss.
+    config_dict : dict
+        Configuration dictionary containing 'z0_field_v_current_G_A' parameter.
+    h : object
+        LakeShore Hall sensor instance.
+    HP1 : object
+        HP1 Power Supply instance with I_read property.
+    """
+    dif_field_G = B0_des_G - h.field_in_G
+    if dif_field_G < 0:
```
atahangarip: Since we either have + or - field with Z_0 shim, if the discrepency is negative, we first decrease the field by 0.8 G to adjust in the intermediate region.

Instruments/field_feedback.py
```diff
@@ -4,6 +4,52 @@
...
+        adjust_field(B0_des_G - 0.8, config_dict, h, HP1)
+    initial_B_field_G = h.field_in_G
+    Z0_initial_current_A = HP1.I_read[0]
+    if HP1.safe_current_on_enable is None:
+        HP1.safe_current_on_enable = 1.5
+    HP1.V_limit[0] = 15.0
+    HP1.I_limit[0] = HP1.round_to_allowed(
+        "I", dif_field_G / config_dict["z0_field_v_current_G_A"]
```
jmfranck: you need to add this to the yaml file that describes the parameters

jmfranck: along with whatever your current estimate of the value is (for it to use as a default)

Instruments/field_feedback.py
```diff
@@ -38,7 +84,7 @@ def adjust_field(B0_des_G, config_dict, h, gen):
     gen.I_limit = I_setting
 
 
-def ramp_field(B0_des_G, config_dict, h, gen):
+def ramp_field(B0_des_G, config_dict, h, gen, HP1):
```
jmfranck, line 152: so your power control server now also needs to create an HP1 instance.

Instruments/field_feedback.py
```diff
@@ -125,21 +171,26 @@ def ramp_field(B0_des_G, config_dict, h, gen):
...
             num_field_matches = 0
-        else:
-            # if it's not within tolerance, and it's not asking for a big
-            # step, then it's asking for an intermediate step
-            logging.info(
-                "You are trying to adjust the field in an intermediate region."
-                "This will be handled by shimstack later. I am leaving field "
-                "as is."
-            )
+        elif field_discrepancy < 0.1:
```
jmfranck: just remove this -- the tolerance_Hz setting above should handle what you're trying to do here.

Instruments/field_feedback.py
```diff
@@ -125,21 +171,26 @@ def ramp_field(B0_des_G, config_dict, h, gen):
...
+                logging.info(
+                    "Field discrepancy is lower than 0.1 G so I am not changing"
+                    "the field!"
+                )
                 break
+        else:
+            # if it's not within tolerance, and it's not asking for a big
+            # step, then it's asking for an intermediate step
+            Z0_adjustment(B0_des_G, config_dict, h, HP1)
+            if abs(h.field_in_G - B0_des_G) < 0.1:
```
jmfranck: here, we just do num_field_matches = 0, because we did have to adjust the field, if only the shims.  Our overall condition here is that we need to hit this if/else block at least twice without having the adjust the field.

Instruments/field_feedback.py
```diff
@@ -125,21 +171,26 @@ def ramp_field(B0_des_G, config_dict, h, gen):
...
+                num_field_matches += 1
+                if num_field_matches > 2:
+                    break
     if num_field_matches < 3:
         raise RuntimeError(
             "I tried 30 times to get my"
-            " field to match within 0.8 G"
+            " field to match within 0.1 G"
```
jmfranck: here replace this with the tolerance number -- give it both in Hz and converted to G.

jmfranck (commented):
comments during meeting

Instruments/power_control_server.py
```diff
@@ -44,89 +45,70 @@ def main():
...
                                             )
-                                            while dBm_setting > last_power + 3:
+                                            if not this_logobj.wg_has_been_flipped:
+                                                # {{{ then I need to turn
+                                                #     everything on
+                                                b.set_wg(True)
+                                                b.set_rf(True)
+                                                b.set_amp(True)
+                                                # }}}
+                                                this_logobj.wg_has_been_flipped = True
```
atahangarip: Couldn't find a way to make this line shorter than 80 chracters

jmfranck (commented):
I rolled back the tolerance to 10 Hz, because 450 Hz is too large.  If we're only asking for 450 Hz, are we using the Z0 at all?

jmfranck (changes requested):
it's inconsistent that the genesys IP is given as a parameter (in the yaml + config) but the prologix IP is not -- do include the prologix IP

jmfranck (commented):
I'm about to merge in more recent changes coming from the hp branch

atahangarip:
> I rolled back the tolerance to 10 Hz, because 450 Hz is too large. If we're only asking for 450 Hz, are we using the Z0 at all?

The resolution in the Z_0 shim is 0.021 G, corresponding to about 90 Hz. I made a small calculation mistake before, which is why it was 450, but I don't think we can do better than 90 Hz.

PR #54, opened 2026-03-18
Y shimming script

atahangarip:
Creating Y shimming solution for correcting field gradients along the sample length.


examples/shim_y.py
```diff
@@ -0,0 +1,291 @@
...
+import SpinCore_pp
+from SpinCore_pp import get_integer_sampling_intervals, save_data
+from SpinCore_pp.ppg import run_spin_echo
+from Instruments import HP6623A, prologix_connection, power_control
+from datetime import datetime
+
+my_exp_type = "ODNP_NMR_comp/Echoes"
+assert os.path.exists(getDATADIR(exp_type=my_exp_type))
+
+# {{{ user settings
```
atahangarip, line 23: I decided to place this parameters here because 1) We may want to change the channel since channel 2 has the worst resolution 2) We may want to wire parallel resistors as we discussed. I thought it would be easier to change these parameters later.

Instruments/shim_y.py
```diff
@@ -0,0 +1,291 @@
...
+my_exp_type = "ODNP_NMR_comp/Echoes"
+assert os.path.exists(getDATADIR(exp_type=my_exp_type))
+
+# {{{ user settings
+Y_channel = 1
+y_current_max = 1.5
+y_voltage_limit = 15.0
+settle_s = 2.0
+skip_field_setting = False
+auto_adc_offset = False
```
atahangarip: I am not sure whether it is best to remove these together with the corresponding if conditions. I thought keeping them wouldn't hurt so that we can do the same test without using any other script if we want.

Instruments/shim_y.py
```diff
@@ -0,0 +1,291 @@
...
+filter_timeconst = 10e-3
+# }}}
+
+
+def fwhm(trace):
```
atahangarip: Since they are long functions I wanted to keep them as a function to make the code more legible - especially for fwhm.

Instruments/shim_y.py
```diff
@@ -0,0 +1,291 @@
...
+    return x[peak_idx], left_cross, right_cross, right_cross - left_cross
+
+
+def integrate_energy(trace, left_lim, right_lim):
+    x = trace.getaxis("t2")
+    y = abs(trace.data) ** 2
+    mask = np.logical_and(x >= left_lim, x <= right_lim)
+    if not mask.any():
+        return 0.0
+    return np.trapz(y[mask], x[mask])
```
atahangarip: I am not sure if this integration is the right one. I will check it but please let me know if you have comment.

examples/shim_y.py
```diff
@@ -0,0 +1,158 @@
...
+
+config_dict = SpinCore_pp.configuration("active.ini")
+
+# {{{ user settings
+Y_channel = config_dict["shim_address"]["Y"][1]
+y_current_max = (
+    1.5  # this is set only in this script, since it doesn't pertain
+)
+#                     to anything else
+y_voltage_limit = 15.0
```
jmfranck: pull this from config dict as well?

Instruments/field_feedback.py
```diff
@@ -183,11 +185,12 @@ def ramp_field(
             elif desired_Z0_current_A > Z0_max_current_A:
                 adjust_main_field(B0_des_G, config_dict, h, gen)
             # }}}
-            HP1.I_limit[config_dict["Z0_channel"]] = HP1.round_to_allowed(
-                "I", 0, desired_Z0_current_A
+            HP1.I_limit[config_dict["shim_address"]["Z0"][1]] = (
+                HP1.round_to_allowed("I", 0, desired_Z0_current_A)
```
jmfranck: check all instances of round_to_allowed -- use the config dict to set the channel.

examples/capture_noise_SC.py
```diff
@@ -133,8 +133,6 @@ def collect(config_dict):
 config_dict = sc.configuration("active.ini")
 # {{{ add file saving parameters to config dict
 config_dict["type"] = "noise" + "_" + str(round(config_dict["SW_kHz"])) + "kHz"
-config_dict["date"] = datetime.now().strftime("%y%m%d")
```
jmfranck, line 136: all of these were moved into save_data

jmfranck: (this came up during my review of your script, which is why this is done here -- essentially I made the change for your example, and propagated the change to all other examples)

PR #55, opened 2026-03-25
Adding shim control to power control server

atahangarip:


examples/set_shim.py
```diff
@@ -0,0 +1,23 @@
...
+        print(
+            "Your optial shim current is "
+            f"{config_dict['shim_y_current_A']} "
+            "and I am setting the field to "
+            f"{shim_current_A}.\n\n"
+            "NOTE: If you want to manually set a particular field "
+            "(including 0 to turn off the shim) you can specify "
+            "as an argument on the command line)"
+        )
+        p.set_shim_current(shim_current_A)
```
jmfranck: this will not currently work -- it requires two arguments.

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
         self.send("SET_FREQ %f" % freq)
         return
 
+    def set_shim_current(self, shim_name, current_A):
```
jmfranck: I would not have both (just to keep it simple)

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
...
+        """Sets the shim current of the specified shim channel.  The shim_name
+        should be one of the keys in the config_dict["shim_channels"] dict, and
+        the current_A should be a float specifying the current in Amps."""
+        self.send("SET_SHIM_CURRENT %s %f" % (shim_name, current_A))
+        retval = self.get()
+        retval = float(retval)
+        return retval
+
+    def set_shim_voltage(self, shim_name, voltage_V):
```
jmfranck: in the plan I'm proposing to just call this `set_shim` to keep it simple.

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
...
+        """Sets the shim voltage of the specified shim channel.  The shim_name
+        should be one of the keys in the config_dict["shim_channels"] dict, and
+        the voltage_V should be a float specifying the voltage in Volts."""
```
jmfranck: needs a proper numpy-style docstring.

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
...
+        self.send("SET_SHIM_VOLTAGE %s %f" % (shim_name, voltage_V))
```
jmfranck: again, proposing just `SET_SHIM`

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
...
+        retval = self.get()
```
jmfranck: what is this??

Instruments/power_control.py
```diff
@@ -158,6 +176,18 @@ def get_field(self):
         retval = float(retval)
         return retval
 
+    def get_shim_current(self, shim_name):
```
jmfranck: I'm proposing in the plan you would have a single `get_shims` that would return a dictionary.

Instruments/power_control_server.py
```diff
@@ -53,7 +53,7 @@ def main():
             address=config_dict["HP1_address"],
         ) as HP1,
     ):
-        HP1.V_limit[config_dict["Z0_channel"]] = 15.0
+        HP1.V_limit[config_dict["shim_address"]["Z0"][1]] = 15.0
```
jmfranck: Rather than having HP1, this would be a `ShimCurrentMapping` instance.  Also, you would replace the Vlimit with an Ilimit.

So, this would be something like `sh_map.I_limit['Z0'] = XXX`

You would also make `safe_current` an arg when initializing `sh_map` b/c you'll need to do that during the `__enter__` method (if you don't know what I'm talking about, read and resolve my edits to `y_shim` first)

Instruments/power_control_server.py
```diff
@@ -179,6 +206,16 @@ def process_cmd(cmd, this_logobj):
                             HP1,
                         )
                         conn.send(("%0.2f" % true_B0_G).encode("ASCII"))
+                    case b"GET_SHIM_CURRENT":
+                        shim_name = args[1].decode("ASCII")
+                        channel = config_dict["shim_channels"][shim_name][1]
+                        current_A = HP1.I_read[channel]
+                        conn.send(("%0.3f" % current_A).encode("ASCII"))
+                    case b"GET_SHIM_VOLTAGE":
```
jmfranck: see other comments -- simpler names, no separate current and voltage.

Instruments/power_control_server.py
jmfranck: since this is at the beginning -- you probably want to add a dict of current limits to your config, and you would set all the voltages to zero to start with.

Instruments/power_control.py
```diff
@@ -134,6 +134,24 @@ def set_freq(self, freq):
         self.send("SET_FREQ %f" % freq)
```
jmfranck, line 141: a not unimportant point is that b/c we've added config stuff to the server, how exactly the server behaves now depends on which directory we call it from (b/c it pulls the active.ini from the current directory).  It's just something to be aware of -- i.e. you want to call from the directory with your local copy of active.ini

Instruments/power_control_server.py
```diff
@@ -72,23 +72,50 @@ def process_cmd(cmd, this_logobj):
...
             print("I split it to ", args)
             if len(args) == 3:
-                if args[0] == b"DIP_LOCK":
-                    freq1 = float(args[1])
-                    freq2 = float(args[2])
-                    _, _, min_f = b.lock_on_dip(
-                        ini_range=(
-                            freq1 * 1e9,
-                            freq2 * 1e9,
+                match args[0]:
```
jmfranck, line 92: the if --> match is I think a good choice.

Instruments/power_control.py
jmfranck: *i.e.* get rid of it?

Instruments/power_control_server.py
jmfranck: note that in general I like the new idea -- where voltages are controlling the strength of the shims, and the current limits are exactly to keep you below your LC-sheet determined limits.

Instruments/power_control_server.py
atahangarip: I kept the separate current and voltage but defined a common function for them as we discussed on Friday

Instruments/power_control_server.py
```diff
@@ -217,6 +247,20 @@ def process_cmd(cmd, this_logobj):
                     case b"GET_FIELD":
                         result = h.field_in_G
                         conn.send(("%0.2f" % result).encode("ASCII"))
+                    case b"GET_SHIM":
+                        retval = (
```
jmfranck, line 267: nice

examples/shim_z0.py
```diff
@@ -0,0 +1,119 @@
...
+            tau_us=config_dict["tau_us"],
+            SW_kHz=config_dict["SW_kHz"],
+            ret_data=data,
+        )
+    # set back to the original voltage at the end
+    p.set_shim_voltage("Z0", orig_voltage_V)
+    print("restored Z0 shim to", orig_voltage_V, "V")
+
+data.rename("indirect", "z0_voltage")
+data.set_axis("z0_voltage", z0_voltage_list).set_units("z0_voltage", "V")
```
jmfranck, line 96: previously, pyspecdata had a stupid thing where `setaxis` didn't follow the usual convention of underscores separating words.  This is just my general note that I've fixed that, and `set_axis` is now supported, and actually preferred (`setaxis` is still supported for backward compatibility, but better to switch to `set_axis` as you write new code)

examples/test_power_control_server.py
```diff
@@ -49,7 +49,6 @@
 log_dict = this_log.log_dict
 logger.debug("log dict:\n" + repr(log_dict))
 with h5py.File("output.h5", "a") as f:
-    log_grp = f.create_group(
-        "log"
+    hdf_save_dict_to_group(
+        f, {"log": this_log.__getstate__()}
```
jmfranck, line 53: notice how this is called in a different way (we create the "log" node here) to accomodate the fixed state rep.

tests/test_logobj.py
```diff
@@ -76,28 +78,52 @@ def build_log(self):
...
-            set(state["array"].keys()),
+            set(state.keys()),
             {"NUMPY_DATA", "dictkeys", "dictvalues"},
         )
         recovered = logobj()
         with tempfile.NamedTemporaryFile(suffix=".h5") as tmpfile:
             with h5py.File(tmpfile.name, "w") as h5file:
-                log_group = h5file.create_group("log")
-                hdf_save_dict_to_group(log_group, state)
+                hdf_save_dict_to_group(h5file, {"log": state})
```
jmfranck, line 98: notice how this saves the log in the same way as the test log write and read examples.  This test passes for me (right now, the socket one above is not passing, but that is unrelated).  If this test is passing, it really seems like you should be able to debug any log problems that you are having.  BUT, I highly suspect problems you have seen come from trying to read in "broken" logs (i.e. ones that are not written using **this** version of the code)

